PR #5, opened 2026-01-28
Implement NameIDMapper as a module to use for translating names to IDs

yushan8:
Translation to OptimizedTargets from name -> id is common functionality. This module is needed for comparing target graphs and converting Results -> `OptimizedTargets` to store the target graph. A separate module `NameIDMapper` is needed to assign names to ID numbers.

Tested with unit test


behinddwalls (approved):
Ship It!

core/targethasher/graph.go
xytan0056: can we have this translation function in another file? To avoid conflicts when we sync Uber targethasher to this file,

yushan8: Good point I'll move it

xytan0056 (approved):

PR #8, opened 2026-02-03
update hasher and fix packages and tests

xytan0056:
- fix up targethasher to be in sync with Uber to include more accurate recursive hash of external repos
- fix unlinked tests in `graph_test.go`
- add Set package


core/targethasher/sourcehasher.go
```diff
@@ -57,69 +47,22 @@ type noOpHasher struct {
...
-func (hh *noOpHasher) HashSourceFile(sourceFile *buildpb.SourceFile, externalHashes map[string]Target) ([]byte, error) {
+func (hh *noOpHasher) HashSourceFile(sourceFile *buildpb.SourceFile) ([]byte, error) {
 	return nil, nil
 }
 
-func (hh *diskHashHelper) HashSourceFile(sourceFile *buildpb.SourceFile, externalHashes map[string]Target) ([]byte, error) {
-	targetName := sourceFile.GetName()
-
-	// If file is excluded, return empty hash
-	for expr := range hh.excludedFiles {
```
xytan0056, line 103: exclusion and fullhashrepo check is moved outside to graph.go

core/common/set/set.go
```diff
@@ -0,0 +1,235 @@
+// Package set implements a generic set of comparable objects.
+// Objects in a set are unordered.
+package set
```
behinddwalls: can we not use something like this? https://github.com/deckarep/golang-set

core/common/set/set.go
```diff
@@ -0,0 +1,235 @@
...
+	return fmt.Sprint(list)
+}
+
+// ToSortedList returns a sorted slice containing all elements of the set. This
+// is a convenience function for sets with elements that can be compared
+// using `<`.
+//
+// ToSortedList does not allocate a slice if the set is empty, returning nil in
+// that case.
+func ToSortedList[T cmp.Ordered](s Set[T]) []T {
```
yushan8: nit: Some of these functions don't seem to be used.

core/common/set/set.go
```diff
@@ -0,0 +1,235 @@
...
+// the set will be empty, but its underlying capacity will be preserved,
+// avoiding the need for a new memory allocation on subsequent insertions.
+// Clear panics if the set is nil.
+func (s Set[T]) Clear() {
+	clear(s)
+}
+
+// Difference builds a new set holding elements present in this set but not in
+// other.
+func (s Set[T]) Difference(other Set[T]) Set[T] {
```
yushan8: nit: also doesn't seem to be used. Maybe we should stick with plain maps since it doesn't seem like we plan on using the methods.

yushan8 (approved):

PR #9, opened 2026-02-04
[tango] Check if attributes and dependencies differ and mark the ChangeType as direct

yushan8:
## Why?
Check if attributes and dependencies differ and mark the ChangeType as direct. This matches the internal behavior.

## What?

1. If the target has a dependency that is a source file, mark the change type as DIRECT. This is faster than going through all the dependencies of the targets and checking for dependency changes.
2. If the dependencies differ, mark the change type as DIRECT.
3. If any of the attributes or their values differ, mark the change type as DIRECT.

## Test Plan

unit test

### DIFF
--- a/core/controller/getchangedtargets.go
+++ b/core/controller/getchangedtargets.go
@@ -276,7 +276,6 @@ func (c *controller) compareTargetGraphs(ctx context.Context, firstGraph, second
 		}
 		if attrsChanged {
 			ct.ChangeType = pb.CHANGE_TYPE_DIRECT
-			continue
 		}
 	}
 
--- a/core/controller/getchangedtargets.go
+++ b/core/controller/getchangedtargets.go
@@ -245,20 +245,38 @@ func (c *controller) compareTargetGraphs(ctx context.Context, firstGraph, second
 		}
 	}
 
-	if sourceFileRuleTypeID != -1 {
-		// Iterate over the changed targets and check if any of them are source files.
-		// If so, check if any of their dependencies are changed source files.
-		// If so, mark the target as a direct change.
-		for name, ct := range changedByName {
-			if ct.GetChangeType() == pb.CHANGE_TYPE_DIRECT {
-				// Already marked as direct, skip
-				continue
-			}
-			newT := secondByName[name]
-			// Ifdependency of the target is a changed source file
-			if hasDepInChangedSourceFileTargets(newT.GetDirectDependencies(), secondMetadata, changedSourceFileTargets) {
-				ct.ChangeType = pb.CHANGE_TYPE_DIRECT
-			}
+	// Iterate over the changed targets and check if any of them are DIRECT changes.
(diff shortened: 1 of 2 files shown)


core/controller/getchangedtargets.go
```diff
@@ -245,20 +245,23 @@ func (c *controller) compareTargetGraphs(ctx context.Context, firstGraph, second
...
+		oldT := firstByName[name]
+
+		// Check if any dependency is a changed source file
+		if hasDepInChangedSourceFileTargets(newT.GetDirectDependencies(), secondMetadata, changedSourceFileTargets) {
+			ct.ChangeType = pb.CHANGE_TYPE_DIRECT
+		// Check if direct dependencies changed.
+		} else if dependenciesChanged(oldT, firstMetadata, newT, secondMetadata) {
+			ct.ChangeType = pb.CHANGE_TYPE_DIRECT
+		} else if attributesChanged(oldT, firstMetadata, newT, secondMetadata) {
+			ct.ChangeType = pb.CHANGE_TYPE_DIRECT
```
xytan0056, line 278: nit: looks like these are OR relationship,
if hasSourceFileCahange OR depCahnge OR attributesChanged {
  ct.ChangeType = DIRECT
}

can be shorter

core/controller/getchangedtargets.go
```diff
@@ -366,6 +369,95 @@ func hasDepInChangedSourceFileTargets(depIds []int32, meta *pb.Metadata, changed
...
+		return true
+	}
+	// Early exit: if both are empty, no change
+	if len(oldDepIDs) == 0 {
+		return false
+	}
+
+	// Cache metadata mappings to avoid repeated map lookups
+	oldTargetIDMapping := oldMeta.GetTargetIdMapping()
+	newTargetIDMapping := newMeta.GetTargetIdMapping()
```
xytan0056, line 411: nit: just in case, let's add an abundance check that `oldTargetIDMapping[oldTarget.ID] == newTargetIDMapping[newTarget.ID]`; return err if not

core/controller/getchangedtargets.go
```diff
@@ -366,6 +369,95 @@ func hasDepInChangedSourceFileTargets(depIds []int32, meta *pb.Metadata, changed
...
+	return false
+}
+
+// attributesChanged checks if the attributes changed between old and new targets.
+func attributesChanged(oldTarget *pb.OptimizedTarget, oldMeta *pb.Metadata, newTarget *pb.OptimizedTarget, newMeta *pb.Metadata) bool {
+	if oldMeta == nil || newMeta == nil {
+		return false
+	}
+
+	oldAttrIDs := oldTarget.GetAttributes()
```
xytan0056, line 442: same here, just in case if some other function wants to call this, we would make sure we're comparing same targets.

xytan0056 (approved):

PR #10, opened 2026-02-04
[tango] Support running get-changed-targets with the cli to test GetChangedTargets

yushan8:
Support running get-changed-targets with cli to test GetChangedTargets against the example server.

### DIFF
--- a/example/tango-config.yaml
+++ b/example/tango-config.yaml
@@ -1,0 +1,8 @@
+remote: "devexp/tango"
+default_branch: "main"
+full_hash_repos:
+  - ""
+exclude_external_targets: true
+bzlmod_enabled: true
+query_timeout: 300
+git_timeout: 300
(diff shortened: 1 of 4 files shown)


xytan0056 (approved):

example/tango-config.yaml
```diff
@@ -0,0 +1,8 @@
+remote: "devexp/tango"
+default_branch: "main"
+full_hash_repos:
+  - ""
+exclude_external_targets: true
+bzlmod_enabled: true
```
xytan0056, line 6: actually we don't respect this flag now.
https://github.com/uber/tango/blob/4f246dc46d2d3b2e2bf0126a0908ad7176c07833/core/bazelrunner/native.go#L59
Still need to figure out a way how to hash external repos in bzlmod ...

yushan8: yeaa...

PR #12, opened 2026-02-09
steam graph responses over chunks

xytan0056:
Modify results to graph response conversion to chunk targets into batches

## Why?

Right now, the graph is written as exactly 2 messages, targets a nd metadata. This is "streamed" back in one large message. 
This PR Change it to write multiple chunked messages. GraphReader already reads message-by-message, so it would also work as-is

## Test Plan
unit test

```
make run-client REMOTE=https://github.com/uber/tango.git  METHOD=get-target-graph BASE_SHA=HEAD
```

### DIFF
--- a/core/common/utils.go
+++ b/core/common/utils.go
@@ -10,6 +10,13 @@ import (
 	"github.com/uber/tango/tangopb"
 )
 
+const (
+	// DefaultTargetChunkSize is the default number of targets per message chunk.
+	// This helps keep individual protobuf messages under size limits and enables
+	// streaming processing of large target graphs.
+	DefaultTargetChunkSize = 1000
+)
+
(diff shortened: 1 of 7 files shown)


yushan8 (approved):

core/common/utils.go
```diff
@@ -10,6 +10,11 @@ import (
 	"github.com/uber/tango/tangopb"
 )
 
+const (
+	// DefaultTargetChunkSize is the default number of targets per message chunk, used for streaming large target graphs.
+	DefaultTargetChunkSize = 1000
```
yushan8, line 15: Should this be configurable?

xytan0056: yes but i think this is too much detail to be left configurable by users. We can expose something in config, but still need a default

PR #15, opened 2026-02-11
implement bzlmod support

xytan0056:
This pr adds initial bzlmod support.

## Why?

in bzlmod, there's no `//external:*` rules, we cannot do "@external_repo//target" => "//external:repo" translation. So we will need to hash the actual external files. Future improvements can include "bazel mod show_repo"

## What?

1. if bzlmod, skip translation
2. `@targets` like (`@gazelle//language/proto:fix.go`) or `@@targets` like (`@@protobuf+//src/google/protobuf/compiler/rust:crate_mapping.cc`) will populate `p.Targets` map
3. recursively hash as if //external rules are not there
4. `@bazel_tools` is a special repo, it comes installed by bazel as `local_repository`, containing  [paltform specific files that don't exist ](https://github.com/bazelbuild/bazel/blob/8ad0dbf9b7fd9ffa520233cc50d8d0a6dc7ecb57/tools/test/BUILD.tools#L142), hashing these will cause `No Such File or Directory`. Ignoring them for now, in future, we should properly hash the rule definition instead

add unit test and fix API call

## Test Plan

unit test 
tested locally
```
make run-server > /tmp/tanx.log 2>&1
make run-client REMOTE=https://github.com/uber/tango.git  METHOD=get-target-graph BASE_SHA=HEAD
```


core/targethasher/graph.go
```diff
@@ -467,11 +475,19 @@ func HashRecursively(ctx context.Context, p HashParam) ([]byte, error) {
 		return t.Hash, nil
 	}
 
+	// bazel_tools is a special case, it comes with bazel installation as a local_repository, it contains platform dependent
+	// files that are not relevant to the build, we don't need to hash it
+	// TODO: use "bazel mod show_repo bazel_tools" to get proper hash
+	if strings.HasPrefix(p.TargetName, "@bazel_tools/") || strings.HasPrefix(p.TargetName, "@@bazel_tools/") {
```
yushan8: Once we update the ExcludedFiles -> ExcludedRegex, should we just add this to the ExcludedRegex list so it's configurable?

yushan8 (approved):
LGTM besides nit

core/targethasher/graph.go
xytan0056: that's an even better idea!

PR #19, opened 2026-02-17
fix changed targets streaming

xytan0056:
Fix streaming bug in `GetChangedTargets` where only the first chunk was read from storage, causing incomplete target graph comparisons.
get-changed-targets call now returns returns nothing and nil metadata

## Why?
The `GetChangedTargets` function was only reading the first chunk from storage, but target graphs are serialized as multiple delimited messages (chunks + metadata). Currently missing metadata entirely. We should do what `GetTargetGraph` does here -- for loop read until eof

## What?

1. properly reads all chunks from the GraphReader until EOF
2. unit tests 
3. misc : cleanup logging

## Test Plan

1. unit tests 
2. local tests
> bazel run //example/client:client -- -addr 127.0.0.1:8081 -method get-changed-targets -remote https://github.com/uber/tango.git -base-sha 47f5e72cf858de87e60ed

### DIFF
--- a/core/controller/getchangedtargets.go
+++ b/core/controller/getchangedtargets.go
@@ -49,10 +49,9 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 	// Start jobs for both revisions. Success or failure, the result will report to the results channel.
 	type graphResult struct {
 		// order is 0 or 1, 0 is the base (first) revision, 1 is the target (second) revision
-		order int
-		// TODO: pb.GetTargetGraphResponse is a stream, so most likely we can't use GetTargetGraphResponse as a return type and we'll want to read it fully before joining the threads
-		graph *pb.GetTargetGraphResponse
-		err   error
+		order  int
+		chunks []*pb.GetTargetGraphResponse
+		err    error
 	}
 	results := make(chan graphResult, len(jobs))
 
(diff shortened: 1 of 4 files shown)


core/controller/getchangedtargets.go
```diff
@@ -66,42 +65,53 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+				if err == io.EOF {
+					break
+				}
+				if err != nil {
+					results <- graphResult{order: idx, err: err}
+					return
+				}
+				chunks = append(chunks, chunk)
 			}
-			graph, err := graphReader.Read()
```
xytan0056, line 77: graphreader only reads one chunk

core/controller/getchangedtargets.go
```diff
@@ -66,42 +65,53 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
-			results <- graphResult{order: idx, graph: graph, err: err}
+			results <- graphResult{order: idx, chunks: chunks}
 		}(i)
 	}
 
 	// Wait for both results to complete, either successfully or with an error.
 	for range jobs {
 		select {
 		case res := <-results:
-			if res.graph != nil {
```
xytan0056, line 86: at this point, the graphreader is done or eof, so we should makr jobs[res.order].completed = true and update jobs[res.order].err accordingly

yushan8 (approved):

PR #22, opened 2026-02-26
implement multi workspace

xytan0056:
Currently repo manager only manages one workspace off of one clone. But we can manage multiple workspaces

## Why?

Multi workspaces clones with "--local" , greatly speeds up the process and increase storage efficiency.

## What?

1. repo mananger manages configurable **pool of workers per repo remote**, each operates on a workspace; one leader for a remote + x workers
2. reuses workers from previous runs via `poolFor()`
3.  if pool is at capacity, Lease blocks until a worker is returned or ctx is cancelled

## Test Plan

CI unit test
local test 
```
make run-server
run //example/client:client -- -addr 127.0.0.1:8081 -method get-changed-targets -remote https://github.com/uber/tango.git -base-sha 47f5e72cf858de87e60eddf58d48b36e70e97082~ -new-base-sha 59932143f27a3e6c230d444f13397f5167db3cef~3  > /tmp/test0.json
```

### DIFF
--- a/core/repomanager/repo_manager.go
+++ b/core/repomanager/repo_manager.go
@@ -38,7 +38,7 @@ type workerPool struct {
 	originMu  sync.Mutex // one lock per repo for orginal clone
 	cloned    bool
 
-	avail chan *workerSlot // available slots; channel size = pool capacity
+	avail chan *workerSlot // available slots; = pool capacity
 }
 
 // workerSlot is a pre-allocated workspace directory that may or may not
--- a/core/repomanager/repo_manager.go
+++ b/core/repomanager/repo_manager.go
@@ -6,21 +6,17 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"time"
+	"sync"
 
 	"github.com/uber/tango/core/git"
 	"github.com/uber/tango/core/workspace"
 	"github.com/uber/tango/tangopb"
-	"github.com/gofrs/flock"
 	"go.uber.org/zap"
 )
 
-const (
-	lockFileName   = ".tango.lease.lock"
-	lockTimeout    = 5 * time.Minute
-	lockRetryDelay = 15 * time.Second
-)
+const defaultPoolSize = 3
 
+// RepoManager manages repository workspaces with a pool of workers per repo.
 type RepoManager interface {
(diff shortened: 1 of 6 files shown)


core/repomanager/repo_manager.go
```diff
@@ -6,21 +6,17 @@ import (
...
 
 	"github.com/uber/tango/core/git"
 	"github.com/uber/tango/core/workspace"
 	"github.com/uber/tango/tangopb"
-	"github.com/gofrs/flock"
 	"go.uber.org/zap"
 )
 
-const (
-	lockFileName   = ".tango.lease.lock"
```
xytan0056, line 19: not needed anymore. We only need a mutex per repo, and workers concurrency is managed by `avail chan`

core/repomanager/repo_manager.go
```diff
@@ -6,21 +6,17 @@ import (
...
-	lockTimeout    = 5 * time.Minute
```
xytan0056, line 20: instead of timeout, we just wait until a workspace is available. If client doesn't wanna wait, it can just cancel the request

core/repomanager/repo_manager.go
```diff
@@ -6,21 +6,17 @@ import (
...
-	lockRetryDelay = 15 * time.Second
-)
+const defaultPoolSize = 3
```
xytan0056, line 17: will do next diff, make it configurable

core/workspace/workspace.go
```diff
@@ -73,10 +69,7 @@ func (w *workspace) Checkout(ctx context.Context, remote string, ref string) err
 	return w.git.Checkout(ctx, ref)
 }
 
-// Release releases the workspace lock.
+// Release is a no-op for the base workspace; pooled workspaces override this.
 func (w *workspace) Release() error {
```
yushan8, line 73: Looks like we don't need this function anymore?

yushan8 (approved):

core/workspace/workspace.go
xytan0056: good cactch! 
It's actually still needed when orchestrator calls `Release` when bazel query and graphing is done. It's just overridden by pooledWorkspce

will do next diff

PR #24, opened 2026-02-26
 reset the workspace before checkout

xytan0056:
before checkout to requested sha , clean up the ws just in case there are leftever files or corruptions

## Test Plan

unit test
local tested

### DIFF
--- a/core/workspace/workspace.go
+++ b/core/workspace/workspace.go
@@ -55,7 +55,6 @@ func (w *workspace) ApplyRequests(ctx context.Context, requests []Request) error
 
 // Checkout resets the workspace to a clean state and checks out the given ref.
 func (w *workspace) Checkout(ctx context.Context, remote string, ref string) error {
-	// Reset any local modifications so the checkout starts clean.
 	if err := w.git.Reset(ctx); err != nil {
 		return fmt.Errorf("git reset --hard: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


core/workspace/workspace.go
```diff
@@ -56,8 +53,15 @@ func (w *workspace) ApplyRequests(ctx context.Context, requests []Request) error
 	return nil
 }
 
-// Checkout checks out the given reference in the workspace.
+// Checkout resets the workspace to a clean state and checks out the given ref.
 func (w *workspace) Checkout(ctx context.Context, remote string, ref string) error {
+	if err := w.git.Reset(ctx); err != nil {
+		return fmt.Errorf("git reset --hard: %w", err)
+	}
+	if err := w.git.Clean(ctx); err != nil {
```
xytan0056, line 64: this potentially can be slow to delete, if there are folders like "bazel-out" or "pkg/mod/" stuff.  But i see current TA does this too. And it's necessary if request is cancel in the middle of git checkout, repo is in a dirty state. We can try this out. 
Another way is to delete the whole WS and ask repo manager to re-clone with `--local`. We can do this if `clean -fdx` is too slow
wdyt? @yushan8

yushan8: I think it'll helpful to benchmark the performance after and see which one is faster with reseting the workspace

yushan8 (approved):

PR #26, opened 2026-02-26
Verify commit passed into the Requests exists in the PR

yushan8:
## Why?

Verify commit passed into the Requests exists in the PR. 
## What?

- Update `gitRequest.Apply` to check if the commit passed in exists in the PR ancestors. If it does not match, reject the commit and return an error.
## Test Plan

unit test

### DIFF
--- a/core/controller/gettargetgraph.go
+++ b/core/controller/gettargetgraph.go
@@ -50,11 +50,6 @@ func (c *controller) getGraph(ctx context.Context, buildDescription *pb.BuildDes
 	if buildDescription.GetBaseSha() == "" || buildDescription.GetRemote() == "" {
 		return nil, fmt.Errorf("build description is missing required fields: base_sha: %s, remote: %s", buildDescription.GetBaseSha(), buildDescription.GetRemote())
 	}
-	for i, req := range buildDescription.GetRequests() {
-		if req.GetCommit() == "" {
-			return nil, fmt.Errorf("request at index %d is missing required commit field (url: %s)", i, req.GetUrl())
-		}
-	}
 	// Look up the the git treehash based on cache path
 	treehashCachePath := common.GetTreehashCachePath(buildDescription)
 	treehashResponse, err := c.storage.Get(ctx, storage.DownloadRequest{Key: treehashCachePath})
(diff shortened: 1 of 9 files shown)


core/workspace/gitrequest.go
```diff
@@ -31,6 +34,13 @@ func (r *gitRequest) Apply(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	actual, err := r.git.RevParse(ctx, fmt.Sprintf("pull/%s/head", r.requestID))
```
xytan0056: what's the difference than `ref`?
```suggestion
	actual, err := r.git.RevParse(ctx, ref)
```

core/workspace/gitrequest.go
```diff
@@ -31,6 +34,13 @@ func (r *gitRequest) Apply(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to resolve PR head: %w", err)
+	}
+	if strings.TrimSpace(actual) != r.commit {
```
xytan0056: it's natural for tango to get requests like
getGraph(pull/123, commit)
getGraph(pull/123, commit^1)so `r.commit` is not always HEAD. Can just check if it's one of ancestors?

core/workspace/gitrequest.go
yushan8: `git fetch origin "+pull/123/head:pull/123/head"` is used for fetching the remote branch and creating a local copy.
The one used in RevParse is checking if the commit exists in the local branch `pull/123/head` after fetching it from the remote

core/workspace/gitrequest.go
yushan8: Ah good point will update

core/controller/gettargetgraph.go
```diff
@@ -50,6 +50,11 @@ func (c *controller) getGraph(ctx context.Context, buildDescription *pb.BuildDes
 	if buildDescription.GetBaseSha() == "" || buildDescription.GetRemote() == "" {
 		return nil, fmt.Errorf("build description is missing required fields: base_sha: %s, remote: %s", buildDescription.GetBaseSha(), buildDescription.GetRemote())
 	}
+	for i, req := range buildDescription.GetRequests() {
+		if req.GetCommit() == "" {
+			return nil, fmt.Errorf("request at index %d is missing required commit field (url: %s)", i, req.GetUrl())
```
xytan0056: SQ wants to allow this case ...
so we can just skip ancestor check if empty
and let's explicitly pass in `{Url: v, Commit:""}` here to indicate
https://github.com/uber/tango/blob/main/example/client/client.go#L59

xytan0056 (approved):

PR #30, opened 2026-02-27
BFS distance from nearest DIRECT change

xytan0056:
## Why?

It's useful to know how impactful is each changed targets. At Uber, if a target is very far away from some file change, it's likely ok to deprioritze testing. So add a distance feature to support it

## What?

implement a BFS from DIRECT changed targets to computes how many hops each changed target is from `CHANGE_TYPE_DIRECT` target in the reverse dependency graph. 
DIRECT targets have distance 0, their parents get 1, and so on. Unreachable targets stay at -1. This should be happening (possible hashing bug).

## Test Plan

unit test
tested locally
image


core/controller/getchangedtargets.go
```diff
@@ -562,6 +565,73 @@ func transposeOptimizedTarget(
...
+				ct.Distance = currentDist + 1
+			}
+		}
+	}
+
+	// Just in case a target is marked changed but has no distance to DIRECT change.
+	// Warn about such cases. Probably a hashing bug.
+	for name, ct := range changedByName {
+		if ct.GetChangeType() == pb.CHANGE_TYPE_INDIRECT && ct.GetDistance() == -1 {
+			logger.Warn("computeDistances: INDIRECT target has no path to a DIRECT change, possible hashing issue",
```
xytan0056, line 628: could have error out, but distance is just nice to have feature, will decide if later people ask for it

yushan8 (approved):

core/controller/getchangedtargets.go
```diff
@@ -289,6 +289,9 @@ func (c *controller) compareTargetGraphs(ctx context.Context, firstGraph, second
 		}
 	}
 
+	// Compute BFS distances from CHANGE_TYPE_DIRECT targets through the dependency graph.
+	computeDistances(c.logger, changedByName, secondByName, secondMetadata)
```
yushan8, line 293: Should we make distance optional? I'm curious about the performance on a large target graph. If there's a degradation, maybe we should have the ability to enable/disable it.

xytan0056: good point. Will do

PR #31, opened 2026-02-27
Update `GetTreehashCachePath` to include computation strategy

yushan8:
## Why?

Since BaseSha could be created as a directory, we also need to prefix `-{COMPUTATION_STRATEGY}` at the end. This ensuresthe filename of the graph stored for BaseSha does not conflict with the directory created.
We also need to sort the encoded RequestURLs to make sure the path returned is consistent.

## What?

- Update `GetTreehashCachePath` to include computation strategy
- Sort the encoded RequestURL strings to make sure the path returned is consistent
## Test Plan

unit test


xytan0056 (approved):
good call

core/common/utils.go
```diff
@@ -29,16 +30,20 @@ func GetGraphByTreeHash(remote, treehash string) string {
...
 func getReqsBase64(requests []*tangopb.Request) string {
+	if len(requests) == 0 {
+		return ""
+	}
 	encodedURLs := make([]string, 0, len(requests))
 	for _, req := range requests {
 		encoded := base64.RawURLEncoding.EncodeToString([]byte(req.GetUrl()))
 		encodedURLs = append(encodedURLs, encoded)
 	}
+	sort.Strings(encodedURLs)
```
xytan0056, line 46: good call

PR #42, opened 2026-03-03
op work for preparing OSS

xytan0056:
operational  work for OSS preparation

## What?

- ISSUE template bug report and feature requests
- CODE_OF_CONDUCT
- CONTRIBUTING guide
- liscense header copy from https://github.com/uber-go/nilaway/blob/main/LICENSE
- dependabot

## Test Plan

CI

### DIFF
--- a/core/bazel/bazel.go
+++ b/core/bazel/bazel.go
@@ -1,3 +1,17 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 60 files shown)


core/bazel/bazel.go
```diff
@@ -1,3 +1,17 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
```
yushan8, line 1: This should be 2026?

yushan8 (approved):

PR #49, opened 2026-03-04
mod tidy files and add corresponding CI check

xytan0056:
- go mod tidy
- remove unused dep for dependabot
- add addtional CI check for lints and dependencies

## Test Plan

CI


yushan8 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
+name: CI check
```
behinddwalls, line 1: can you create a gate check like this : https://github.com/uber/submitqueue/blob/main/.github/workflows/ci.yml#L116-L133 make sure if anything fails, CI fails and blocks the merge

behinddwalls (changes requested):

.github/workflows/ci.yml
xytan0056: i have set up branch protection that requires all CI check to pass

behinddwalls: Screenshot 2026-03-03 at 7 25 42 PM


Not really this only does for build and test and it becomes a bit of pain to update in UI vs just have a gate in the  CI.yml itself, https://github.com/uber/tango/settings/rules/12154016

behinddwalls: Also branch protection things are deprecated, you need to update the RuleSets

behinddwalls: https://github.com/uber/tango/settings/branch_protection_rules/73552968 i see you have this, you may want to just add to RulSet itself

behinddwalls (approved):

.github/workflows/ci.yml
xytan0056: gotcha. Thanks for the tips

Issue #28, opened 2026-02-27
context cancellation not propagated


xytan0056:
to repro, `make server`, then
```
bazel run //example/client:client -- -addr 127.0.0.1:8081 -method get-changed-targets -remote https://github.com/uber/tango.git -base-sha cf3f54d15c652854f31fb552183cfb26ecde1813 -new-base-sha 59932143f27a3e6c230d444f13397f5167db3cef~1  > /tmp/t.json
```
then ctrl+C, the query is still running on server, not cancelled

PR #58, opened 2026-03-07
feat(cancel): Graceful cancellation to GetChangedTargets

sbalabanov:
Proper cancellation from GRPC context


xytan0056:
fixes  #28

could you test if this fixes it?

sbalabanov:
the context should be passed down properly anyways

xytan0056 (approved):

PR #60, opened 2026-03-16
[tango] Cache GetChangedTargets results

### LINKED ISSUE #59: cache GetChangedTargets results
We can further improve cache hit rate by caching the changed targets between target graphs via treehash

yushan8:
## Why?

To optimize GetChangedTargets and reduce redundant computation, we will implement a caching layer keyed by the tree hashes of Revision 1 and Revision 2. By persisting these results to storage, we can bypass recalculation for repeated requests, significantly improving response latency and cache hit rates.
Also fix export file in build.

## What?

GetChangedTargets implementation:
1. Ensure tree hashes for both revisions are present in storage.
2. Return the cached result if available.
3. Otherwise, calculate changed targets between Revision 1 and 2.
4. Cache the results by uploading them to storage.
## Test Plan

unit test
## Issue

fixes #59

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -21,6 +21,7 @@ import (
 	"io"
 
 	"github.com/uber/tango/core/common"
+	"github.com/uber/tango/core/storage"
 	pb "github.com/uber/tango/tangopb"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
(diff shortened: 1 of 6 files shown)


controller/getchangedtargets.go
```diff
@@ -695,3 +757,18 @@ func getDefaultDistance(outputConfig *pb.OutputConfig, forNewTarget bool) int32
 	}
 	return -1
 }
+
+// readTreehash fetches the treehash stored at GetTreehashCachePath for the given build description.
+// Returns an empty string on any error or cache miss so callers can treat it as an optional optimistic lookup.
+func readTreehash(ctx context.Context, st storage.Storage, buildDescription *pb.BuildDescription) string {
```
xytan0056, line 763: this is a good util function to use globally
https://github.com/search?q=repo%3Auber%2Ftango+common.GetTreehashCachePath&type=code

xytan0056 (approved):

controller/getchangedtargets.go
yushan8: good point will update in follow up PR and refactor any areas that uses it

PR #70, opened 2026-03-31
Cache GetChangedTargetsAndEdges

yushan8:
Cache GetChangedTargetsAndEdges so subsequent requests to the endpoint is fast.
Tested by running twice to see that ` GetChangedTargetsAndEdges` is cached between the two graphs.
```
bazel run //example/client -- --method=get-changed-targets-and-edges --remote=https://github.com/uber/tango.git --base-sha "f7dfb7e240de0a0e509c4f343f6ea6c81cd8f9fd" --new-base-sha "cdff82ac070ce15cfa5f5f957492889741c2898b"
```

### DIFF
--- a/controller/getchangedtargetsandedges.go
+++ b/controller/getchangedtargetsandedges.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/uber/tango/core/common"
+	"github.com/uber/tango/core/storage"
 	pb "github.com/uber/tango/tangopb"
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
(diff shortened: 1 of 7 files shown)


xytan0056 (approved):

core/common/utils.go
```diff
@@ -56,6 +56,13 @@ func GetComparedTargetsCachePath(remote, treehash1, treehash2 string) string {
 	return filepath.Join("compared-targets", ToShortRemote(remote), treehash1, treehash2)
 }
 
+// GetComparedTargetsAndEdgesCachePath returns the cache path for a GetChangedTargetsAndEdges result.
+// treehash1 and treehash2 are the resolved treehashes of the first and second revisions.
+// remote is the shared git remote for both revisions.
+func GetComparedTargetsAndEdgesCachePath(remote, treehash1, treehash2 string) string {
+	return filepath.Join("compared-targets-and-edges", ToShortRemote(remote), treehash1, treehash2)
```
xytan0056, line 63: should we treat `(treehash1, treehash2)` `(treehash2, treehash1)` the same?

xytan0056, line 63:
```suggestion
// GetComparedTargetsAndEdgesCachePath returns the cache path for a GetChangedTargetsAndEdges result.
// treehash1 and treehash2 are the resolved treehashes of the first and second revisions.
// remote is the shared git remote for both revisions.
func GetChangedTargetsAndEdgesCachePath(remote, treehash1, treehash2 string) string {
	return filepath.Join("changed-targets-and-edges", ToShortRemote(remote), treehash1, treehash2)
```

controller/getchangedtargetsandedges.go
```diff
@@ -46,6 +47,56 @@ func (c *controller) GetChangedTargetsAndEdges(request *pb.GetChangedTargetsAndE
...
+	cacheStart := time.Now()
+	treehash1 := readTreehash(ctx, c.storage, request.GetFirstRevision())
+	treehash2 := readTreehash(ctx, c.storage, request.GetSecondRevision())
+	if treehash1 != "" && treehash2 != "" {
+		cacheKey := common.GetComparedTargetsAndEdgesCachePath(request.GetFirstRevision().GetRemote(), treehash1, treehash2)
+		cachedReader, cacheErr := storage.NewChangedTargetsAndEdgesReader(ctx, c.storage, cacheKey)
+		if cacheErr != nil && !storage.IsNotFound(cacheErr) {
+			c.logger.Warn("GetChangedTargetsAndEdges: Failed to read from cache, proceeding to compute", zap.Error(cacheErr))
+		} else if cachedReader != nil {
+			var cached []*pb.GetChangedTargetsAndEdgesResponse
```
xytan0056, line 60: log some here about the cache hit and send a `count` metric wrt the hit too

core/common/utils.go
yushan8: hmm no cause I think treehash1 would usually be the old revision and treehash2 is the new revision. That means the return values of `GetChangedTargetsAndEdges` are different i.e. a target that's added in `(treehash1, treehash2)` is considered deleted in `(treehash2, treehash1)`

controller/getchangedtargetsandedges.go
yushan8: yes good point, I'll add metrics for all the cache hit in follow up PR

PR #72, opened 2026-04-02
[bazel] Add an optional param to bazel client to include `--profile`

yushan8:
Add an optional param to bazel client to include `--profile`.

### DIFF
--- a/config/repository_config.go
+++ b/config/repository_config.go
@@ -24,4 +24,5 @@ type RepositoryConfig struct {
 	BzlmodEnabled          bool     `yaml:"bzlmod_enabled"`
 	BazelCommand           string   `yaml:"bazel_command"`
 	QueryTimeout           int64    `yaml:"query_timeout"` // in seconds
+	BazelExtraArgs         []string `yaml:"bazel_extra_args"`
 }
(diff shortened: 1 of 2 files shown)


core/bazel/bazel.go
```diff
@@ -56,6 +56,7 @@ type BazelClient struct {
 	logger             *zap.SugaredLogger
 	execCommandContext func(ctx context.Context, name string, arg ...string) commander
 	queryTimeout       time.Duration
+	profilePath        string
```
xytan0056: profile path is per query call, not per bazel instance.
so i think is should only be one of `additionalArgs` to query. 
And the value should be a temp file created by query caller.
This behavior should be enabled in repo config. 
```
if profileEnabled:
     profileF:= os.CreateTemp()
     additionalArgs = append(additionalArgs, "--profile="+profileF)
     bazel.Query(... addtionalArgs)
     storage.Upload(profileF)
```

yushan8: Synced offline. Add bazel_extra_args in repo_config instead.

xytan0056 (approved):

PR #98, opened 2026-06-02
Add more granular metrics on failed requests

yushan8:
Add more granular metrics for different failure categories.

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -121,7 +121,7 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 					)
 					scope.Counter("cache_hit").Inc(1)
 					scope.Timer("cache_read_duration").Record(cacheReadDuration)
-				if sendErr := sendWithDistanceFilter(stream, cached, maxDist); sendErr != nil {
+					if sendErr := sendWithDistanceFilter(stream, cached, maxDist); sendErr != nil {
 						logger.Error("GetChangedTargets: Failed to send cached response", zap.Error(sendErr))
 						failureReason = "send"
 						return fmt.Errorf("failed to send cached response: %w", sendErr)
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -66,7 +66,7 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		failureReason = "validation"
 		return err
 	}
-	scope = scope.Tagged(map[string]string{"repo": request.GetFirstRevision().GetRemote()})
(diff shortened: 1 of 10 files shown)


controller/getchangedtargets.go
```diff
@@ -41,17 +41,33 @@ type job struct {
...
+	failureReason := ""
 	defer func() {
 		if retErr != nil {
 			scope.Counter("failure").Inc(1)
+			reason := failureReason
+			if reason == "" {
+				reason = classifyError(retErr)
+			}
+			ft := "user"
+			if reason != "validation" && reason != "cancelled" && reason != "deadline_exceeded" {
```
justinwon777: the reason strings are reused throughout the PR. Maybe we should keep a list of constants in errors.go

controller/getchangedtargets.go
```diff
@@ -41,17 +41,33 @@ type job struct {
...
+				ft = "infra"
```
justinwon777: same with this. We should have constants for user and infra

controller/getchangedtargets.go
```diff
@@ -244,6 +262,7 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 	secondGraph = nil
 	if err != nil {
 		logger.Error("GetChangedTargets: Failed to compare target graphs", zap.Error(err))
+		failureReason = "compare"
 		return fmt.Errorf("failed to compare target graphs: %w", err)
```
justinwon777: Should we define custom error types in errors.go, so that in `classifyError` it can just match the error type instead of looking for strings. I believe that's the best practice for our internal services. Then maybe all the classification can happen inside `classifyError` instead of doing some in there and some like this.

controller/getchangedtargets.go
yushan8: Yup good point!

controller/getchangedtargets.go
yushan8: Yes good point. Created an interface for ClassifyError. That way we can categorize failures internally.

orchestrator/native_orchestrator.go
```diff
@@ -75,17 +75,17 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, param GetTarget
 	cfg, err := config.Parse(b.configFilePath)
 	if err != nil {
 		logger.Errorw("GetTargetGraph: Error parsing config file", zap.String("configFilePath", b.configFilePath), zap.Error(err))
-		return nil, err
+		return nil, common.WithReason("config_parse", common.ErrorTypeInfra, err)
```
justinwon777: could create a list of const in this file for the failure reasons like in controller/errors.go

justinwon777 (approved):

PR #102, opened 2026-06-08
Parallelize treehash reads GetChangedTargets

sywhang:
Parallelize treehash reads GetChangedTargets
Fetch both revision treehashes concurrently with a WaitGroup
instead of of sequentially, saving one storage RTT on every
cache-eligible request.

Also replace the queue[0]/queue[1:] BFS pattern with an index
loop so the backing array is not stranded in memory during traversal.

Pre-allocate the BFS queue with len(changedByName) capacity to reduce early growth cycles.

Added a few benchmark tests which confirms ~2x speedup as expected.

### DIFF
--- a/controller/benchmark_test.go
+++ b/controller/benchmark_test.go
@@ -1,0 +1,157 @@
+// Copyright (c) 2026 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


xytan0056 (approved):

sbalabanov:
`readTreehash` swallows errors and it should not.
With the right error processing, the implementation of parallelization will be slightly more complex.

PR #103, opened 2026-06-08
feat(cancel): Thread ctx into compare helpers for graceful cancellation

sbalabanov:
## Summary
- Extends graceful cancellation through the shared compare pipeline so a client disconnect terminates long-running graph diffs quickly.
- `getTargetsAndMetadata`, `buildNameIndex`, `computeDistances` now take `ctx` and return `error`; helpers check ctx at chunk boundaries or every `cancelCheckInterval` iterations of their inner loops.
- `compareTargetGraphs` and `compareTargetGraphsAndEdges` thread ctx into the helpers and propagate cancellation errors. `compareTargetGraphsAndEdges` also adds periodic ctx checks inside its big inline loops; redundant boundary checks dropped now that helpers and inline loops check ctx themselves.

## Test plan
- [x] `go build ./...`
- [x] `go test ./controller/...`
- [ ] CI green

## Note
Stacks on top of #58 (which adds the `ctx` param to `compareTargetGraphs` too). If #58 lands first the trivial signature overlap will need a rebase here.

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -303,29 +303,26 @@ func (c *controller) compareTargetGraphs(ctx context.Context, logger *zap.Logger
 
 	// 1) Extract targets and metadata; index by canonical names
 	indexStart := time.Now()
-	firstTargetsByID, firstMetadata := getTargetsAndMetadata(firstGraph)
-
-	if ctx.Err() != nil {
-		return nil, ctx.Err()
+	firstTargetsByID, firstMetadata, err := getTargetsAndMetadata(ctx, firstGraph)
+	if err != nil {
+		return nil, err
 	}
-
-	secondTargetsByID, secondMetadata := getTargetsAndMetadata(secondGraph)
-
-	if ctx.Err() != nil {
-		return nil, ctx.Err()
+	secondTargetsByID, secondMetadata, err := getTargetsAndMetadata(ctx, secondGraph)
+	if err != nil {
+		return nil, err
 	}
-
 	// Release raw chunk slices — individual target protos are now held by the ID maps.
 	firstGraph = nil
 	secondGraph = nil
-	firstByName := buildNameIndex(firstTargetsByID, firstMetadata)
-	firstTargetsByID = nil // all pointers are now in firstByName; drop the duplicate map
-
-	if ctx.Err() != nil {
-		return nil, ctx.Err()
+	firstByName, err := buildNameIndex(ctx, firstTargetsByID, firstMetadata)
+	if err != nil {
+		return nil, err
 	}
-
-	secondByName := buildNameIndex(secondTargetsByID, secondMetadata)
+	firstTargetsByID = nil // all pointers are now in firstByName; drop the duplicate map
+	secondByName, err := buildNameIndex(ctx, secondTargetsByID, secondMetadata)
+	if err != nil {
+		return nil, err
+	}
 	secondTargetsByID = nil
 	indexDuration := time.Since(indexStart)
 	scope.Timer("index_duration").Record(indexDuration)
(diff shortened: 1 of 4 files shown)


controller/getchangedtargetsandedges.go
sbalabanov, line 227: this becomes not fully correct with the change
a better generalized error classification framework is needed

controller/getchangedtargets.go
```diff
@@ -523,21 +543,26 @@ func getTargetsAndMetadata(graph []*pb.GetTargetGraphResponse) (map[int32]*pb.Op
 			}
 		}
 	}
-	return targets, merged
+	return targets, merged, nil
 }
 
 // buildNameIndex creates name->target maps using the provided metadata information.
-func buildNameIndex(targetsByID map[int32]*pb.OptimizedTarget, meta *pb.Metadata) map[string]*pb.OptimizedTarget {
+func buildNameIndex(ctx context.Context, targetsByID map[int32]*pb.OptimizedTarget, meta *pb.Metadata) (map[string]*pb.OptimizedTarget, error) {
```
xytan0056, line 580: this is just building metadata, ctx isn't really needed here

controller/getchangedtargets.go
```diff
@@ -781,16 +806,21 @@ func sendWithDistanceFilter(stream pb.TangoServiceGetChangedTargetsYARPCServer,
 // enqueued, so they keep their initial distance of -1 (out-of-range).
 //
 // Targets unreachable from any DIRECT target keep the initial distance of -1.
-func computeDistances(logger *zap.Logger, changedByName map[string]*pb.ChangedTarget, targetsByName map[string]*pb.OptimizedTarget, meta *pb.Metadata, maxDistance int32) {
+func computeDistances(ctx context.Context, logger *zap.Logger, changedByName map[string]*pb.ChangedTarget, targetsByName map[string]*pb.OptimizedTarget, meta *pb.Metadata, maxDistance int32) error {
 	if meta == nil {
```
xytan0056, line 840: same here. I think if ctx is cancelled when computing distances, we should just let it go, otherwise for the same pattern we're gonna have to check context for every operation

controller/getchangedtargets.go
```diff
@@ -286,21 +286,33 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 	return nil
 }
 
-func (c *controller) compareTargetGraphs(logger *zap.Logger, firstGraph, secondGraph []*pb.GetTargetGraphResponse, maxDist int32, outputDistances bool) ([]*pb.GetChangedTargetsResponse, error) {
+func (c *controller) compareTargetGraphs(ctx context.Context, logger *zap.Logger, firstGraph, secondGraph []*pb.GetTargetGraphResponse, maxDist int32, outputDistances bool) ([]*pb.GetChangedTargetsResponse, error) {
```
xytan0056, line 299: I think ctx cancel saves most resources when its I/O call or holding resources like files, db queries, when it's pure in-mem, these finish in microseconds/milliseconds, adding ctx just adds noise to the signature

controller/getchangedtargets.go
sbalabanov: Context was added based on the analysis of whether a code path is long running assuming monorepo size. If it iterates over millions of records, it is guarded by cancellation check. A well behaved program should gracefully finish subsecond.

controller/getchangedtargets.go
sbalabanov: for long running operations we'll have to.
not for every operation, but sampled.
https://github.com/uber/tango/pull/108 is the follow up

controller/getchangedtargets.go
sbalabanov: >when it's pure in-mem, these finish in microseconds/milliseconds
not for compute-intense operations

xytan0056 (approved):

PR #105, opened 2026-06-08
Replace multierr with stdlib errors.Join

sbalabanov:
## Summary
- Replace `go.uber.org/multierr` usages with stdlib `errors.Join` (Go 1.20+)
- Drop the direct `multierr` dependency from `go.mod`; it remains as an indirect dependency of `go.uber.org/fx`

Three call sites updated:
- `core/itg/changeanalyzer/changeanalyzer.go` — combine diff + fetch errors
- `controller/getchangedtargets.go` — accumulate per-job errors into a slice and join
- `controller/getchangedtargetsandedges.go` — same pattern as above

## Test plan
- [x] `go build ./...`
- [x] `go test ./core/itg/changeanalyzer/... ./controller/...`
- [x] `go test ./...` (one pre-existing, unrelated failure in `core/targethasher` — missing `NewMockSourceHasher`, reproduces on `main`)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -40,7 +40,6 @@ use_repo(
     "org_uber_go_fx",
     "org_uber_go_goleak",
     "org_uber_go_mock",
-    "org_uber_go_multierr",
     "org_uber_go_yarpc",
     "org_uber_go_zap",
 )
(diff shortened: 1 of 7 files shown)


xytan0056:
we need `go.uber.org/fx` regardless, because when this is imported into Uber, we need to provide uber's buildkite orchestrator as fx module

sbalabanov:
> we need `go.uber.org/fx` regardless, because when this is imported into Uber, we need to provide uber's buildkite orchestrator as fx module

multierr remains as a transitive dependency, but there is no use for it now with latest Go versions

xytan0056 (approved):

PR #108, opened 2026-06-08
Harden context propagation across long-running paths

sbalabanov:
## Summary

Six related fixes so cancellation (client disconnect, ctx timeout) actually stops the work instead of letting it run to completion in the background. Audit found these via a codebase sweep for long-running loops and I/O without ctx checks.

1. **bazelisk download** — `ensureBazelisk` uses `http.NewRequestWithContext` so the download aborts on cancel. `NewBazelClient` and `detectBazelExecutable` now take a `ctx`.
2. **Cache-drain loops** — `GetChangedTargets` and `GetChangedTargetsAndEdges` check `stream.Context().Err()` each iteration so we stop pulling cached chunks if the client gave up.
3. **bazel query stream parser** — `getQueryResult` polls `ctx.Err()` every `cancelCheckInterval` messages while unmarshalling.
4. **`ResultToGetTargetGraphResponse`** — takes `ctx` and polls every `cancelCheckInterval` targets in the per-target loop.
5. **Storage Put** — disk and memory storage wrap the upload reader in a small `storage.CtxReader` so `io.Copy` returns `ctx.Err()` promptly on cancel instead of blocking.
6. **targethasher loops** — filter loops in `GetInternalTargetsWithoutHashAndRootInfo` / `HashExternalTargets` poll every `cancelCheckInterval` entries; root/cycle/sequential dispatcher loops in `fromProto` check once per outer iteration since each `HashRecursively` call can be long.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./...` passes (only pre-existing `targethasher` failure is bazel-only mock generation)
- [x] `bazel test //core/targethasher:targethasher_test` passes
- [x] `bazel build` of all affected packages clean
- [ ] CI green

### DIFF
--- a/core/storage/disk/disk.go
+++ b/core/storage/disk/disk.go
@@ -79,7 +79,7 @@ func (d *diskStorage) Put(ctx context.Context, req storage.UploadRequest) error 
 	tmpPath := tmp.Name()
 	defer os.Remove(tmpPath)
 
-	if _, err := io.Copy(tmp, req.Reader); err != nil {
+	if _, err := io.Copy(tmp, &storage.CtxReader{Ctx: ctx, R: req.Reader}); err != nil {
 		tmp.Close()
 		return err
 	}
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -92,6 +92,11 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 				var cached []*pb.GetChangedTargetsResponse
 				var readErr error
 				for {
+					if err := ctx.Err(); err != nil {
+						cachedReader.Close()
+						// Client gave up while we were draining the cache. Surface as a user-cancelled error.
+						return common.WithReason(failureReasonCancelled, common.ErrorTypeUser, err)
+					}
 					var resp *pb.GetChangedTargetsResponse
 					resp, readErr = cachedReader.Read()
 					if readErr == io.EOF {
--- a/core/bazel/bazel.go
+++ b/core/bazel/bazel.go
@@ -67,7 +67,7 @@ type Params struct {
 	QueryTimeout       time.Duration
 }
 
-func NewBazelClient(p Params) (*BazelClient, error) {
+func NewBazelClient(ctx context.Context, p Params) (*BazelClient, error) {
 	execCmd := p.ExecCommandContext
 	if execCmd == nil {
 		execCmd = func(ctx context.Context, name string, arg ...string) commander {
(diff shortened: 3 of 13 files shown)


sywhang:
@sbalabanov I think you have to sign the CLA still 😂

core/targethasher/graph.go
```diff
@@ -51,6 +51,11 @@ const (
 	GeneratedFileType = "generated file"
 
 	UnknownRuleType = "unknown rule"
+
+	// cancelCheckInterval is how often we poll ctx.Err() inside per-target hot loops.
+	// Picked to keep overhead negligible while still surfacing cancellation in <100ms
+	// for typical target rates.
+	cancelCheckInterval = 4096
```
xytan0056: i agree with the other IO bounded context checks  
But the `cancelCheckInterval` is a bit overdone imo. tehse are pure inmemory loops (filtering, proto fields, map lookups)even with 100k targets the whole loop finishes well under 100ms, so very little benefit to warrant a mid-loop cancellation, especially given the code complexity it adds: the fact that the same `cancelCheckInterval` and `if i%cancelCheckInterval == 0` check is sprinkled in multiple places

core/targethasher/graph.go
```diff
@@ -334,7 +349,11 @@ func HashExternalTargets(ctx context.Context, r *buildpb.QueryResult, targets ma
 		ExcludedRegex: nil,
 		UseBzlmod:     useBzlmod,
 	}
+	// No sampling: each HashRecursively call is heavy, so per-iteration check is free relative to body.
 	for _, name := range externalRuleNames {
+		if err := ctx.Err(); err != nil {
```
xytan0056: `HashRecursively` already accepts `ctx` If that is cancelled, the next recursive call will surface it internally. The check right before calling is redundant.

core/storage/disk/disk.go
```diff
@@ -79,7 +79,7 @@ func (d *diskStorage) Put(ctx context.Context, req storage.UploadRequest) error
 	tmpPath := tmp.Name()
 	defer os.Remove(tmpPath)
 
-	if _, err := io.Copy(tmp, req.Reader); err != nil {
+	if _, err := io.Copy(tmp, &storage.CtxReader{Ctx: ctx, R: req.Reader}); err != nil {
```
xytan0056, line 82: agree with the intention but btw we have a reason to NOT propogate client cancellation:
if a client cancels, existing graph computation should continue and written to storage, otherwise given sq other clients' frequent cancellation, tango would be doing a lot of wasted overhead work.

sbalabanov:
> @sbalabanov I think you have to sign the CLA still 😂

https://support.github.com/ticket/personal/0/4446946

core/storage/disk/disk.go
sbalabanov: So, if the client cancels, we do not cancel and keep the connection?
If we need async job to always complete (also hesitant on this one - will the cancelled request ever be reissued?), then we need a separate infra to synchronize the GRPC stream and job done, rather than delaying GRPC request not reacting to cancellation.

sbalabanov:
I removed checks where deemed as not needed, but left it for the storage writes

xytan0056 (approved):

PR #115, opened 2026-06-12
Pass app-lifetime context to controller, orchestrator, repomanager

sbalabanov:
## Summary

Streaming handlers in the controller only watched the gRPC stream context, so an in-flight request kept running until the client disconnected even after the server began shutting down. This PR adds an app-lifetime context so handlers — and the async cache-write goroutine — can also be cancelled at server teardown.

### Changes

1. **All three top-level constructors now accept an app-lifetime context as their first argument:**
   - `controller.NewController(appCtx, Params)`
   - `orchestrator.NewNativeOrchestrator(appCtx, Params)`
   - `repomanager.NewRepoManager(appCtx, Params)`

2. **`controller.linkRequestCtx(reqCtx)` helper** returns a context cancelled by either the request's stream context (client disconnect) or the controller's `appCtx` (server shutdown), backed by `context.AfterFunc` so the watcher handle is released by the deferred cancel — `appCtx` itself is never cancelled by this helper. Each streaming handler (`GetChangedTargets`, `GetTargetGraph`) now derives its working context from this helper and threads it through all downstream calls.

3. **Goroutines launched from handlers inherit the linked context.** The per-job graph-fetch workers in `GetChangedTargets` are derived via `context.WithCancel(ctx)`, so they abort on either client disconnect or app shutdown. The async cache-write goroutine instead passes `c.appCtx` directly to storage — the write must survive client disconnect (so detaching from the request context is the whole point) but must NOT outlive the server, and `c.appCtx` satisfies both with no extra plumbing. Per-operation deadlines are the storage backend's responsibility.

4. **\`example/main.go\`** wires \`appCtx\` from \`signal.NotifyContext(ctx, SIGINT, SIGTERM)\` and propagates it to all three constructors.

## Test plan
- [x] \`go build ./...\`
- [x] \`go test ./controller/... ./orchestrator/... ./core/repomanager/...\`
- [x] New tests cover both cancellation paths (request ctx, app ctx) and the AfterFunc-release semantics, including an end-to-end test that drives \`GetChangedTargets\` and asserts the cache-write goroutine's context is cancelled by \`appCtx\` only, not by request cancellation.
- [ ] CI green

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -56,7 +56,8 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		return common.WithReason(failureReasonValidation, common.ErrorTypeUser, err)
 	}
 	scope = scope.Tagged(map[string]string{"repo": common.ToShortRemote(request.GetFirstRevision().GetRemote())})
-	ctx := stream.Context()
+	ctx, cancelLink := c.linkRequestCtx(stream.Context())
+	defer cancelLink()
 	start := time.Now()
 	logger := c.logger.With(
 		zap.Any("first_revision", request.GetFirstRevision()),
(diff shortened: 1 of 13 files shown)


controller/getchangedtargets.go
```diff
@@ -269,17 +270,18 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
-		treehash1 := readTreehash(cacheCtx, c.storage, logger, request.GetFirstRevision())
-		treehash2 := readTreehash(cacheCtx, c.storage, logger, request.GetSecondRevision())
+		// Use c.appCtx directly: the cache write is fire-and-forget and must
+		// outlive the request (so a client disconnect doesn't abort it) but
+		// must NOT outlive the server (so it doesn't leak past shutdown).
+		// c.appCtx fits both: it's never cancelled by client disconnect and
+		// is cancelled on shutdown. Per-operation deadlines are the storage
+		// backend's responsibility — the controller is backend-agnostic and
+		// must not encode any one implementation's I/O budget.
+		treehash1 := readTreehash(c.appCtx, c.storage, logger, request.GetFirstRevision())
```
xytan0056, line 283: the context here should be detached from existing one, so that when c.appctx is cancelled, the upload still continues on best effort

sbalabanov: App context is cancelled when application shuts down as it is signalled by the host management platform. SIGKILL is very likely to follow soon after. How should "best effort" look like in this case?

xytan0056: I'm thinking best effort of uploading results to survive both client disconnect and app shutdown -- nothing cancels it, upload whatever it can.

xytan0056 (approved):

controller/getchangedtargets.go
sbalabanov: there are few severe anti-patterns in this decision as I see of.
https://dave.cheney.net/2016/12/22/never-start-a-goroutine-without-knowing-how-it-will-stop

PR #116, opened 2026-06-12
Honor OutputConfig include_hashes/include_tags/include_attributes

xytan0056:
The three include_* flags on OutputConfig are documented but never read: nothing in the controller calls GetIncludeHashes/GetIncludeTags/ GetIncludeAttributes, so omitting OutputConfig (proto3 default: all false) still returned fully-populated targets.

Apply the flags at send time, matching the existing max_distance pattern — cache continues to store the fully-populated payload, and each request shapes its own output by shallow-copying targets with the stripped fields cleared. No new cache-key dimensions; one cache entry serves any combo.

Wired in GetChangedTargets (sendWithDistanceFilter) and GetTargetGraph (per-chunk filter before stream.Send). GetChangedTargetGraph is still a stub. The shallow copy never mutates the source, so the cache-write goroutine and send loop remain race-free.

Behavior change: existing clients that omit OutputConfig will now receive slim responses. They must set IncludeHashes / IncludeTags / IncludeAttributes to true to keep the old payload shape.

## Test Plan

uni test

tested locally 
```
make run-client-changed-targets REMOTE=https://github.com/uber/tango.git BASE_SHA=872881fd~1 NEW_BASE_SHA=872881fd INCLUDE_ATTRIBUTES=true > /tmp/tt.json
```
verified  only attrs are added in response


controller/getchangedtargets.go
```diff
@@ -855,15 +855,21 @@ func transposeOptimizedTarget(
...
 // sendWithDistanceFilter streams responses to the client, filtering changed targets to those
-// within maxDist from any distance-0 seed when maxDist >= 0.
-// Metadata and other non-target responses are always forwarded.
-// Filtering and sending are combined into a single pass to avoid an intermediate allocation.
-func sendWithDistanceFilter(stream pb.TangoServiceGetChangedTargetsYARPCServer, responses []*pb.GetChangedTargetsResponse, maxDist int32) error {
+// within maxDist from any distance-0 seed when maxDist >= 0, and stripping per-target
+// hash/tags/attributes per outputConfig's include_* flags. Metadata and other non-target
+// responses are always forwarded. Filtering and sending are combined into a single pass
+// to avoid an intermediate allocation.
+func sendWithDistanceFilter(stream pb.TangoServiceGetChangedTargetsYARPCServer, responses []*pb.GetChangedTargetsResponse, maxDist int32, outputConfig *pb.OutputConfig) error {
```
yushan8: Should we just rename this to `sendWithOutputConfigFilters` since it's doing more filtering beyond distance?

yushan8 (approved):

PR #117, opened 2026-06-15
Detach cancellation in cache-write goroutine; log infra errors

xytan0056:
The compared-targets cache-write goroutine in GetChangedTargets used context.Background(), losing the request's tracing/identity baggage. Switch to context.WithoutCancel(ctx) so the write survives client disconnect while keeping request values.
Per-operation deadlines stay out of the controller — storage.Storage is backend-agnostic and must not encode any one backend's I/O budget. Backends that need a deadline (e.g. YARPC/terrablob, which rejects deadline-less contexts with InvalidArgument: missing TTL) own that responsibility in their Put/Get implementations.

readTreehash now takes a *zap.Logger and logs any non-NotFound storage error. The silent swallow is what hid this bug — a genuine cache miss stays quiet, but an infra failure disabling the cache is now visible. The goroutine also warns when it skips because a treehash was empty.

## Test Plan
Unit tests
tested in Uber's internal env

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -74,12 +74,12 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 	}
 
 	// Try to serve from cache first using the stored treehashes for both revisions.
-	// readTreehash returns "" on any miss/error so we silently skip the cache when
-	// either treehash is not yet available.
+	// readTreehash returns "" on a cache miss (and logs any real storage error) so
+	// we skip the cache when either treehash is not yet available.
 	if !request.GetBypassCache() {
 		cacheStart := time.Now()
-		treehash1 := readTreehash(ctx, c.storage, request.GetFirstRevision())
-		treehash2 := readTreehash(ctx, c.storage, request.GetSecondRevision())
+		treehash1 := readTreehash(ctx, c.storage, logger, request.GetFirstRevision())
+		treehash2 := readTreehash(ctx, c.storage, logger, request.GetSecondRevision())
 		if treehash1 != "" && treehash2 != "" {
 			cacheKey := common.GetComparedTargetsCachePath(request.GetFirstRevision().GetRemote(), treehash1, treehash2, request.GetRequestOptions())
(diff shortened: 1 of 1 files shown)


controller/getchangedtargets.go
```diff
@@ -1007,14 +1018,21 @@ func validateGetChangedTargetsRequest(request *pb.GetChangedTargetsRequest) erro
 
 // readTreehash fetches the treehash stored at GetTreehashCachePath for the given build description.
 // Returns an empty string on any error or cache miss so callers can treat it as an optional optimistic lookup.
-func readTreehash(ctx context.Context, st storage.Storage, buildDescription *pb.BuildDescription) string {
-	resp, err := st.Get(ctx, storage.DownloadRequest{Key: common.GetTreehashCachePath(buildDescription)})
+// A genuine cache miss (not-found) is silent; any other storage error is logged so an
+// infra failure that disables the cache (e.g. a missing-deadline "missing TTL" reject) is visible.
+func readTreehash(ctx context.Context, st storage.Storage, logger *zap.Logger, buildDescription *pb.BuildDescription) string {
```
xytan0056, line 1021: add a log here for verbosity

controller/getchangedtargets.go
```diff
@@ -266,14 +266,25 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
 	go func() {
-		cacheCtx := context.Background()
-		treehash1 := readTreehash(cacheCtx, c.storage, request.GetFirstRevision())
-		treehash2 := readTreehash(cacheCtx, c.storage, request.GetSecondRevision())
+		// terrablob calls go over YARPC, which derives the request TTL from the
+		// context deadline; a context without one is rejected with
+		// "InvalidArgument: missing TTL". Detach cancellation so the cache write
+		// survives client disconnect, but preserve the request values and set a
+		// fresh deadline so the storage reads/writes carry a valid TTL.
+		const cacheWriteTimeout = 2 * time.Minute
```
yushan8: Can this be configurable in the service config?

xytan0056: removed it and decided to add deadline in storage implementation

yushan8 (approved):

PR #125, opened 2026-06-23
classify getGraph errors and consolidate failure reasons

justinwon777:
Consolidate shared failure reason constants (cancelled, unknown, storage, validation, deadline_exceeded) into core/common so they can be reused across controller, orchestrator, and external orchestrator implementations.

Classify previously-unclassified error paths in getGraph:
- context cancellation during graph read → cancelled/user
- context cancellation during orchestrator call → cancelled/user
- graph blob NotFound after treehash hit → falls through to compute
- non-NotFound storage errors → graph_fetch/infra
- unclassified orchestrator errors → graph_fetch/infra

## Test Plan

unit tests


controller/getchangedtargets.go
xytan0056, line 255: need to classify this? like "orchestrator error"?

xytan0056: I'm thinking what error would we get if buildkite fails

xytan0056 (approved):

PR #127, opened 2026-06-29
Clean up logging and remove double-logged errors

yushan8:
Removed double-logged errors (logged then returned — caller logs again):
- core/workspace/gitrequest.go — 5 Errorw calls in Apply() (fetch, diff, patch, commit, submodule). The orchestrator already logs when ApplyRequests fails.
- core/bazel/bazel.go — Errorw for bazel executable detection. Caller logs.
- core/bazel/query.go — Errorf in wrapQueryFailure. The error message already carries stderr; logging it separately was redundant.
- orchestrator/native_orchestrator.go — 11 Errorw calls across the entire GetTargetGraph method. All errors are returned as ClassifiedError and the controller boundary emits failure metrics + logs.

Corrected log levels:
- controller/getchangedtargets.go:55 — validation failure Error → removed (linter removed the log entirely since it's returned with classification)
- controller/getchangedtargets.go:293 — background goroutine cache-write skip Error → Warn (not a service failure, can't return)
- controller/gettargetgraph.go:107 — treehash cache miss Info → Debug (normal hot-path noise)
- core/bazel/bazel.go:94 — constructor detail Info → Debugw
- core/bazel/query.go:40 — bazel query detail Infow → Debugw

Fixed unstructured logging:
- orchestrator/native_orchestrator.go:137 — Errorf → Errorw with zap.Error()
- core/bazel/query.go:103 — Debugf → Debugw with zap.Int()

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -52,7 +52,7 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		}
 	}()
 	if err := validateGetChangedTargetsRequest(request); err != nil {
-		c.logger.Error("GetChangedTargets: Invalid request", zap.Error(err))
+		c.logger.Warn("GetChangedTargets: Invalid request", zap.Error(err))
 		return common.WithReason(common.FailureReasonValidation, common.ErrorTypeUser, err)
 	}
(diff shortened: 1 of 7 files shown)


sbalabanov (approved):
in many cases it is beneficial to wrap the error with `fmt.Errorf` to include additional metadata and preserve call stack in the error chain.

PR #132, opened 2026-06-30
Clean up logging in controller and wrap errors

yushan8:
- Replaced common.WithReason with fmt.Errorf for error wrapping in getchangedtargets.go and getGraph.
- Promoted silent failures to hard errors — cache read failures, corrupt blobs, and cachedReader.Close() errors now return instead of logging and falling through.
- Surfaced ws.Release() errors in native_orchestrator.go via multierr.Append instead of swallowing them.
- Added logger.Error at each error return in top-level controller (GetTargetGraph, GetChangedTargets) with specific failure messages.
- Removed non-actionable logs — cache miss debug logs, bypass_cache logs, context cancelled logs.
- Removed unused logger param from compareTargetGraphs.
- Updated tests to match new error wrapping and removed ClassifiedError assertions where common.WithReason was replaced.

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -43,7 +43,6 @@ type job struct {
 func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, stream pb.TangoServiceGetChangedTargetsYARPCServer) (retErr error) {
 	scope := c.scope.SubScope("get_changed_targets")
 	scope.Counter("calls").Inc(1)
-	logger := c.logger
 	defer func() {
 		if retErr != nil {
 			scope.Counter("failure").Inc(1)
@@ -53,14 +52,14 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		}
 	}()
 	if err := validateGetChangedTargetsRequest(request); err != nil {
-		logger.Error("GetChangedTargets: validation failed", zap.Error(err))
+		c.logger.Error("GetChangedTargets: validation failed", zap.Error(err))
 		return fmt.Errorf("validate request: %w", err)
 	}
 	scope = scope.Tagged(map[string]string{"repo": common.ToShortRemote(request.GetFirstRevision().GetRemote())})
 	ctx, cancelLink := c.linkRequestCtx(stream.Context())
 	defer cancelLink()
 	start := time.Now()
-	logger = c.logger.With(
+	logger := c.logger.With(
 		zap.Any("first_revision", request.GetFirstRevision()),
 		zap.Any("second_revision", request.GetSecondRevision()),
 	)
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -43,6 +43,7 @@ type job struct {
 func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, stream pb.TangoServiceGetChangedTargetsYARPCServer) (retErr error) {
 	scope := c.scope.SubScope("get_changed_targets")
 	scope.Counter("calls").Inc(1)
+	logger := c.logger
 	defer func() {
 		if retErr != nil {
 			scope.Counter("failure").Inc(1)
@@ -52,18 +53,17 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		}
 	}()
 	if err := validateGetChangedTargetsRequest(request); err != nil {
-		return common.WithReason(common.FailureReasonValidation, common.ErrorTypeUser, err)
+		logger.Error("GetChangedTargets: validation failed", zap.Error(err))
+		return fmt.Errorf("validate request: %w", err)
 	}
 	scope = scope.Tagged(map[string]string{"repo": common.ToShortRemote(request.GetFirstRevision().GetRemote())})
 	ctx, cancelLink := c.linkRequestCtx(stream.Context())
 	defer cancelLink()
 	start := time.Now()
-	logger := c.logger.With(
+	logger = c.logger.With(
 		zap.Any("first_revision", request.GetFirstRevision()),
 		zap.Any("second_revision", request.GetSecondRevision()),
 	)
-
-	logger.Info("GetChangedTargets: Processing request")
 
 	// Default max_distance to -1 (no filtering) when the client omits OutputConfig
 	// entirely. When OutputConfig is supplied, take max_distance at face value —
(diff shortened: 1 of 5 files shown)


controller/getchangedtargets.go
```diff
@@ -52,19 +53,18 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		}
 	}()
 	if err := validateGetChangedTargetsRequest(request); err != nil {
-		return common.WithReason(common.FailureReasonValidation, common.ErrorTypeUser, err)
+		logger.Error("GetChangedTargets: validation failed", zap.Error(err))
```
sbalabanov: it does quite the opposite of what PR description says

controller/getchangedtargets.go
```diff
@@ -98,9 +99,10 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 				var readErr error
 				for {
 					if err := ctx.Err(); err != nil {
-						cachedReader.Close()
-						// Client gave up while we were draining the cache. Surface as a user-cancelled error.
-						return common.WithReason(common.FailureReasonCancelled, common.ErrorTypeUser, err)
+						if closeErr := cachedReader.Close(); closeErr != nil {
+							return fmt.Errorf("close cached reader: %w (original: %w)", closeErr, err)
+						}
```
sbalabanov: timeout error should prevail, but reader close error can be combined into it

orchestrator/BUILD.bazel
```diff
@@ -20,6 +20,7 @@ go_library(
         "//graphrunner",
         "//tangopb",
         "@com_github_uber_go_tally//:tally",
+        "@org_uber_go_multierr//:multierr",
```
sbalabanov: https://github.com/uber/tango/pull/105

sbalabanov (changes requested):

controller/getchangedtargets.go
yushan8: Oops, yea just updated it. All errors are logged once at the controller level.

controller/getchangedtargets.go
```diff
@@ -82,14 +82,15 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+			logger.Error("GetChangedTargets: failed to read revision treehash", zap.Error(err))
+			return fmt.Errorf("read revision treehash: %w", err)
 		}
 		if treehash1 != "" && treehash2 != "" {
 			cacheKey := common.GetComparedTargetsCachePath(request.GetFirstRevision().GetRemote(), treehash1, treehash2, request.GetRequestOptions())
 			cachedReader, cacheErr := storage.NewChangedTargetsReader(ctx, c.storage, cacheKey)
 			if cacheErr != nil && !storage.IsNotFound(cacheErr) {
-				logger.Warn("GetChangedTargets: Failed to read from cache, proceeding to compute", zap.Error(cacheErr))
+				logger.Error("GetChangedTargets: failed to read from cache", zap.Error(cacheErr))
+				return fmt.Errorf("failed to read from cache: %w", cacheErr)
```
xytan0056: I think we should still fall through and continue to compute in this case if storage read errors out

controller/getchangedtargets.go
```diff
@@ -113,30 +115,26 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
-					}
-					totalDuration := time.Since(start)
-					logger.Info("GetChangedTargets: Successfully streamed from cache",
-						zap.Duration("total_duration", totalDuration),
-					)
-					scope.Timer("total_duration").Record(totalDuration)
-					scope.Histogram("total_duration.histogram", c.totalDurationBuckets).RecordDuration(totalDuration)
-					return nil
+					logger.Error("GetChangedTargets: failed to read cached response", zap.Error(readErr))
+					return fmt.Errorf("read cached response: %w", readErr)
```
xytan0056: same here, any kind of transient storage blip that used to digrade gracefully now hard fails

orchestrator/native_orchestrator.go
```diff
@@ -128,19 +127,17 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, param GetTarget
 		return nil, fmt.Errorf("lease workspace: %w", err)
 	}
 	defer func() {
-		err := ws.Release()
-		if err != nil {
-			// clean up the workspace if release fails.
+		if releaseErr := ws.Release(); releaseErr != nil {
```
xytan0056, line 131: This too. workspace Release shoudn't fail the request when the compute result is already put in storage.

controller/getchangedtargets.go
yushan8: Yea I think if there's an actual storage failure, it'll be caught in getGraph anyways.

controller/getchangedtargets.go
```diff
@@ -52,7 +52,8 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		}
 	}()
 	if err := validateGetChangedTargetsRequest(request); err != nil {
-		return common.WithReason(common.FailureReasonValidation, common.ErrorTypeUser, err)
+		c.logger.Error("GetChangedTargets: validation failed", zap.Error(err))
```
sywhang, line 55: please don't log *and* return the errors. Suppose you have a chain of funcs each returning an error:

func1 -> func2 -> func3

and you log before returning the error each time. 

You'll end up with logs that look like:

"func3: some error happened"
"func2: error while doing something: func3: some error happened"
"func1: another thing: func2: error while doing something: func3: some error happened"

This is completely redundant and likely not what you want.

controller/getchangedtargets.go
```diff
@@ -82,14 +81,14 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		cacheStart := time.Now()
 		treehash1, treehash2, err := readTreehashParallel(ctx, c.storage, request.GetFirstRevision(), request.GetSecondRevision())
 		if err != nil {
-			logger.Error("GetChangedTargets: Failed to read revision treehash", zap.Error(err))
-			return common.WithReason(failureReasonTreehashRead, common.ErrorTypeInfra, err)
+			logger.Error("GetChangedTargets: failed to read revision treehash", zap.Error(err))
```
sywhang, line 84: same issue here

orchestrator/native_orchestrator.go
sywhang: Q: does workspace know how to recover itself if this fails?

controller/getchangedtargets.go
yushan8: All errors are logged once at the controller level now. The rest of the functions errors logs were removed https://github.com/uber/tango/pull/127. Shouldn't we have service logs for errors at this level?

PR #133, opened 2026-06-30
CI: add coverage step

sywhang:
This adds coverage step in the CI by leveraging codecov.

Right now the config isn't going to block on any specific numbers yet, but it should at least allow us to collect coverage on every PR so we know when we're not adding appropriate tests.


sywhang:
There's a legit test failure on main. Gotta fix that.

sbalabanov (approved):

PR #135, opened 2026-06-30
core/bazel: fix race

sywhang:
Fix a race during Bazel query cancellation by replacing concurrent pipe readers with file-backed stdout/stderr capture. 

This avoids buffer races and inherited-pipe hangs while preserving live stderr, context cancellation, SIGTERM/SIGKILL escalation, and context-aware parsing. 

Added regression coverage for cancellation and descendant processes holding output descriptors.


yushan8 (approved):

core/bazel/stream.go
```diff
@@ -32,6 +32,7 @@ func streamOutput(ctx context.Context, src io.Reader, dst io.Writer) error {
 
 	select {
 	case <-ctx.Done():
+		<-done
```
sbalabanov: this blocks on copying goroutine, in which case it does not make any sense to have a goroutine at all but rather block on a main thread.
what we need to have instead is a cancellable Copy, or at least a cancellable io.Reader.
This wrapper is better than nothing on pure blocking reads: https://github.com/uber/tango/pull/108/changes#diff-279a703791b2ff1b50139ea545c8aca00928b154a414330229e4891ec895aaf2

sbalabanov (changes requested):

core/bazel/stream.go
sywhang: the main thread is already blocked before we call errgroup.Wait() because we're calling cmd.Wait().

PR #137, opened 2026-07-01
fix golangci-lint detected issues

sywhang:
This fixes all issues discovered while enabling golangci-lint in this repo.

This includes all checks included by the default golangci-lint configuration, including but not limited to:
- staticcheck
- errcheck
- ineffassign
- dead code

Will follow up with another PR to enable the CI check for this.


core/targethasher/mock_sourcehasher_test.go
```diff
@@ -0,0 +1,56 @@
+// Code generated by MockGen. DO NOT EDIT.
```
xytan0056: this file is generated by bazel
https://github.com/uber/tango/blob/4765546c1e0e3165f19b3f2e57d13f06351d322d/core/targethasher/BUILD.bazel#L43

controller/getchangedtargets.go
```diff
@@ -257,8 +260,8 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 	compareStart := time.Now()
 	changedTargetsResponses, err := c.compareTargetGraphs(ctx, logger, firstGraph, secondGraph, maxDist)
 	// Allow GC of raw graph data while the caching goroutine runs.
-	firstGraph = nil
-	secondGraph = nil
+	firstGraph = nil  //nolint:ineffassign // intentional: allow GC of large slice
+	secondGraph = nil //nolint:ineffassign // intentional: allow GC of large slice
```
sbalabanov: now sure why this is needed?

core/bazel/stream.go
```diff
@@ -24,38 +24,28 @@ import (
...
-		_, err := io.Copy(dst, src)
-		done <- err
-	}()
-
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case err := <-done:
-		return err
+	buf := make([]byte, 32*1024)
```
sbalabanov: should be separate change?

core/bazel/stream.go
```diff
@@ -24,38 +24,28 @@ import (
...
+	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		n, readErr := src.Read(buf)
```
sbalabanov: ctxReader works ok with io.Copy, and no need to reimplement io.Copy then

sbalabanov (changes requested):
PR likely combined two independent changes.

controller/getchangedtargets.go
sywhang: ineffassign detects this as not a strictly necessarily assignment since assigning nil to this doesn't change its semantics.

core/bazel/stream.go
sywhang: yeah, remnant from a different branch. removed

xytan0056 (approved):

PR #143, opened 2026-07-06
Converge CancelCheckInterval into a common constant

yushan8:
Converge CancelCheckInterval into a common constant. Move service related constants to service_config.go.

### DIFF
--- a/core/common/utils.go
+++ b/core/common/utils.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	buildpb "github.com/bazelbuild/buildtools/build_proto"
+	"github.com/uber/tango/config"
 	tgerrors "github.com/uber/tango/core/errors"
 	"github.com/uber/tango/core/targethasher"
 	"github.com/uber/tango/tangopb"
@@ -202,14 +203,14 @@ func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Res
 	attrStrValIDToVal := attrStrValMapper.Invert()
 
 	// chunk targets into multiple messages for streaming
-	responses := chunkTargets(optimizedTargets, tgerrors.DefaultTargetChunkSize)
+	responses := chunkTargets(optimizedTargets, config.DefaultTargetChunkSize)
 	for _, meta := range ChunkMetadata(
 		targetIDToName,
 		ruleTypeIDToName,
 		tagIDToName,
 		attrNameIDToName,
 		attrStrValIDToVal,
-		tgerrors.DefaultMetadataMapChunkSize,
+		config.DefaultMetadataMapChunkSize,
 	) {
 		responses = append(responses, &tangopb.GetTargetGraphResponse{
 			Item: &tangopb.GetTargetGraphResponse_Metadata{Metadata: meta},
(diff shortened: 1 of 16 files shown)


core/targethasher/BUILD.bazel
```diff
@@ -5,11 +5,14 @@ go_library(
...
     srcs = [
         "graph.go",
+        "response.go",
         "sourcehasher.go",
     ],
     importpath = "github.com/uber/tango/core/targethasher",
     visibility = ["//visibility:public"],
     deps = [
+        "//core/common",
+        "//tangopb",
```
sbalabanov: well now targethasher needs a knowledge of protobuf (which is a UX/App layer of onion). Is it a good idea?

sbalabanov (changes requested):

core/common/utils.go
```diff
@@ -15,20 +15,21 @@
...
-	buildpb "github.com/bazelbuild/buildtools/build_proto"
-	"github.com/uber/tango/core/targethasher"
 	"github.com/uber/tango/tangopb"
 )
 
 const (
+	// CancelCheckInterval is how often we poll ctx.Err() inside per-target hot loops.
+	// Picked to keep overhead negligible while still surfacing cancellation in <100ms
+	// for typical target rates.
+	CancelCheckInterval = 4096
```
sywhang: common is now becoming a super thick package with all kinds of random stuff in it, and exported. 

Let's not do this.

core/targethasher/BUILD.bazel
yushan8: Removed. Will move this to the mapper directory in subsequent PR.

core/common/utils.go
yushan8: Yea good point. Moved the constant to the errors package instead. Will move the mapper functions to subsequent PR.

core/errors/errors.go
```diff
@@ -0,0 +1,8 @@
+package errors
+
+const (
+	// CancelCheckInterval is how often we poll ctx.Err() inside per-target hot loops.
+	// Picked to keep overhead negligible while still surfacing cancellation in <100ms
+	// for typical target rates.
+	CancelCheckInterval = 4096
```
liujsj: Prior we were using 1024 with this comment. Should we not converge on the smaller of the two? The cost of doing the check is almost negligible?

core/errors/constants.go
```diff
@@ -0,0 +1,8 @@
+package errors
```
liujsj, line 1: Is this really an error package? Should it be some constant package instead

core/common/utils.go
```diff
@@ -224,14 +203,14 @@ func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Res
 	attrStrValIDToVal := attrStrValMapper.Invert()
 
 	// chunk targets into multiple messages for streaming
-	responses := chunkTargets(optimizedTargets, DefaultTargetChunkSize)
+	responses := chunkTargets(optimizedTargets, config.DefaultTargetChunkSize)
```
liujsj: Why does this not use the service config value? If someone specified in their config, then it would get ignored here and instead use the default value still.

core/errors/constants.go
yushan8: I don't think constant package is a common practice. I think usually the constants are closely related to what the package/functions are doing. In this case, we're polling to see if the context was cancelled (error) during long running loops. But yea I'll update the file to constants.go.

core/common/utils.go
yushan8: Ah good catch

PR #146, opened 2026-07-06
[core/workspace] fix silent drop of invalid scheme

sywhang:
core/workspace's NewRequest API returns a nil, nil on an invalid scheme. Instead we should return an error when we encounter a value we do not understand.

### DIFF
--- a/core/workspace/request_test.go
+++ b/core/workspace/request_test.go
@@ -63,5 +63,6 @@ func TestNewRequest_InvalidScheme(t *testing.T) {
 	rawURL := "phabricator://bad"
 
 	_, err := NewRequest(rawURL, g, "baseRef", "", zap.NewNop().Sugar())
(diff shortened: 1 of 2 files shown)


core/workspace/request_test.go
```diff
@@ -45,3 +58,10 @@ func TestNewRequest_InvalidURL(t *testing.T) {
 	require.Error(t, err)
 	require.Nil(t, req)
 }
+
+func TestNewRequest_InvalidScheme(t *testing.T) {
+	rawURL := "phabricator://bad"
+
+	_, err := NewRequest(rawURL, g, "baseRef", "", zap.NewNop().Sugar())
+	require.ErrorContains(t, err, "unsupported scheme: phabricator")
```
sbalabanov: plz do not assert on error message; claude.md should have an instruction around this

sbalabanov (approved):
plz fix test assert before landing

PR #149, opened 2026-07-07
core/storage: stream graph writes instead of buffering in full

sywhang:
WriteGraphStream and WriteChangedTargetsStream marshaled the entire response set into a bytes.Buffer before Put, keeping a second full copy of the serialized graph resident in memory. Replace this with a shared generic writeStream helper that encodes length-delimited protobuf into an io.Pipe while Put consumes the reader, so the payload is never fully buffered a second time.

The writer goroutine checks ctx before each message so a cancellation unwinds the encode loop promptly and propagates the context error to the reader; if Put returns first, the reader is closed to unblock the goroutine. Both writers now delegate to the helper, dropping the redundant bytes.NewReader(buf.Bytes()) wrapper.

There's another writer we can dedup using this, and should be fairly straightforward since writeStream uses generics but that will be a follow up.

### DIFF
--- a/core/storage/graphwriter.go
+++ b/core/storage/graphwriter.go
@@ -31,10 +31,12 @@ import (
 // The writer goroutine checks ctx before each message so a cancellation
 // unwinds the encode loop promptly instead of waiting for Put to notice and
 // stop reading; the context error is propagated to the reader. If Put returns
-// before draining the pipe, the reader is closed to unblock (and terminate)
-// the writer goroutine either way.
+// before draining the pipe, the reader is closed to unblock the writer. The
+// goroutine is joined before returning, and its error is returned when Put
+// succeeds.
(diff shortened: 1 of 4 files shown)


core/storage/graphwriter.go
```diff
@@ -15,24 +15,49 @@
...
+			if err = w.WriteMsg(m); err != nil {
+				err = fmt.Errorf("write delimited: %w", err)
+				break
+			}
+		}
+		pw.CloseWithError(err)
+	}()
+	err := st.Put(ctx, UploadRequest{Key: key, Reader: pr})
+	// Unblock the writer goroutine if Put stopped reading early.
+	pr.CloseWithError(err)
```
sbalabanov: ideally have to wait for goroutine to complete in this case before proceeding further to avoid dangling routines

sbalabanov (approved):

xytan0056 (approved):

PR #152, opened 2026-07-07
[core/itg/graph] fix nil err return case propagation

sywhang:
invalidateHashRecursively never fails even though it returns an error in its siganture, as in it always returns nil.

But since this has an error return type, its callers are propagating that nil error everywhere and also nil checking that error.

This cleans that up.


core/itg/graph/invalidate.go
sywhang, line 107: actually, this should be returned. abandoning this PR and will follow up with a different one.

PR #155, opened 2026-07-07
[core/itg/graph] Fix invalidation error not propagated

sywhang:
In invalidateHashRecursively we never return an error even though the function signature supports returning an error and its callsites check for errors.

When the target is not found or if the target's hash is nil, it silently swallows and continues the loop. Skipping that causes us to actually skip all of its rdeps too.

This shouldn't be the case - we should return the error when it's not found at all. If the Hash is nil, then we check if it's already in our visited set (`invalidated`) instead of just silently skipping over it.

### DIFF
--- a/core/itg/graph/invalidate.go
+++ b/core/itg/graph/invalidate.go
@@ -105,7 +105,7 @@ func (g *OptimizedGraph) invalidateHashRecursively(ids IntSet, invalidated IntSe
 		if !ok {
 			return fmt.Errorf("target %d not found", targetID)
 		}
-		if target.Hash == nil {
+		if invalidated.Contains(targetID) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


core/itg/graph/invalidate.go
```diff
@@ -102,7 +102,10 @@ func (g *OptimizedGraph) invalidateHashRecursively(ids IntSet, invalidated IntSe
 		targetID := candidates[curIdx]
 		curIdx++
 		target, ok := g.OptimizedTargets[targetID]
-		if !ok || target.Hash == nil {
+		if !ok {
+			return fmt.Errorf("target %d not found", targetID)
```
xytan0056, line 106: this is intentional.
if the target doesn't exist (such as external rule targets naturally doesn't exits in OptimizedTargets). There's nothing to invalidate in ITG graph. 

what's actually erroring out is `removeTarget` below.

PR #156, opened 2026-07-07
[tango]: Add integration testing

JacobOaks:
Add some integration testing to tango. These tests:

* Spin up the tango server, configured with the tango repo itself as the target
* Create a client connected to it
* Call its APIs (GetTargetGraph and GetChangedTargets)
* Asserts some responses from the server against expected data for actual commits in the tango repository.

I decided to use the actual tango repository to provide a heightened realism to the tests.
I also prototype a setup that generated an entire repository with multiple commits at runtime,
but just the boilerplate to bring in rules_go/other basics to support any non-contrived examples
seems to complicate the test code enough that I figured I might as well just use tango itself.
The downside of this approach is that I couldn't easily really assert against any actual piece of the graph
in entirety in testing GetTargetGraph, because it's pretty huge in a real repository with external dependencies. Instead, I assert against some specific expected edges.

This provides a useful signal for how well tango works as an entire system. In a future PR, we can also re-use the same testing components to write benchmarks that measure tango's latency.

This also adds a separate make target and CI step to run this since, in local testing, it can take about a minute or so.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -19,17 +19,18 @@ package integration_test
 
 import (
 	"context"
-	"fmt"
 	"io"
 	"net"
 	"os"
 	"path/filepath"
 	"sort"
 	"testing"
+	"text/template"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/uber/tango/config"
 	"github.com/uber/tango/controller"
 	"github.com/uber/tango/core/git"
 	"github.com/uber/tango/core/repomanager"
@@ -42,7 +43,10 @@ import (
 	"go.uber.org/zap/zaptest"
 )
 
-const requestTimeout = 10 * time.Minute
+const (
+	requestTimeout     = 10 * time.Minute
+	configTemplateFile = "testdata/tango-config.yaml.tmpl"
+)
 
 func repoRemote(t *testing.T) string {
 	t.Helper()
@@ -53,32 +57,33 @@ func repoRemote(t *testing.T) string {
 
 func writeConfig(t *testing.T, dir, remote, clonePath, workerPath string) string {
 	t.Helper()
+
+	tmpl, err := template.ParseFiles(configTemplateFile)
+	require.NoError(t, err, "failed to parse config template")
+
 	configPath := filepath.Join(dir, "tango-config.yaml")
-	content := fmt.Sprintf(`storage:
-  type: "memory"
+	f, err := os.Create(configPath)
+	require.NoError(t, err, "failed to create config file")
+	defer f.Close()
 
-repository:
-  - remote: %q
-    default_branch: "main"
-    full_hash_repos: [""]
-    excluded_files: ["^@@?bazel_tools/"]
-    exclude_external_targets: true
-    bzlmod_enabled: true
-    query_timeout: 600
+	err = tmpl.Execute(f, struct {
+		Remote     string
+		ClonePath  string
+		WorkerPath string
+	}{
+		Remote:     remote,
+		ClonePath:  clonePath,
+		WorkerPath: workerPath,
+	})
+	require.NoError(t, err, "failed to render config template")
 
-service:
-  worker_pool_size: 2
-  repo_manager_clone_path: %q
-  worker_root_path: %q
-`, remote, clonePath, workerPath)
-	require.NoError(t, os.WriteFile(configPath, []byte(content), 0o644))
 	return configPath
 }
 
 func startServer(t *testing.T, remote string) string {
 	t.Helper()
 
-	cacheDir := filepath.Join(os.TempDir(), "tango-e2e-cache")
+	cacheDir := filepath.Join(t.TempDir(), "tango-e2e-cache")
 	require.NoError(t, os.MkdirAll(cacheDir, 0o755))
 	t.Setenv("XDG_CACHE_HOME", cacheDir)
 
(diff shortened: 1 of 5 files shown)


integration/integration_test.go
```diff
@@ -0,0 +1,476 @@
...
+	remote := os.Getenv("TANGO_REPO_REMOTE")
+	require.NotEmpty(t, remote, "TANGO_REPO_REMOTE must be set (pass --test_env=TANGO_REPO_REMOTE=... to bazel test)")
+	return remote
+}
+
+func writeConfig(t *testing.T, dir, remote, clonePath, workerPath string) string {
+	t.Helper()
+	configPath := filepath.Join(dir, "tango-config.yaml")
+	content := fmt.Sprintf(`storage:
+  type: "memory"
```
xytan0056: let's make this a `data` to `go_test` and read from there

integration/integration_test.go
```diff
@@ -0,0 +1,476 @@
...
+  worker_root_path: %q
+`, remote, clonePath, workerPath)
+	require.NoError(t, os.WriteFile(configPath, []byte(content), 0o644))
+	return configPath
+}
+
+func startServer(t *testing.T, remote string) string {
+	t.Helper()
+
+	cacheDir := filepath.Join(os.TempDir(), "tango-e2e-cache")
```
xytan0056:
```suggestion
	cacheDir := filepath.Join(t.TempDir(), "tango-e2e-cache")
```

integration/integration_test.go
```diff
@@ -0,0 +1,476 @@
...
+		assert.Equal(t, int32(3), ct.Distances["//example:example"])
+	})
+
+	t.Run("new_targets", func(t *testing.T) {
+		// Compare two adjacent commits:
+		//   046de2c (parent)
+		//   1f2e3e9 Honor OutputConfig include_hashes/include_tags/include_attributes (#116)
+		// The second commit adds controller/output_filter.go and output_filter_test.go.
+		const firstSHA = "046de2c20b5492cd5606d32fd632a38b8b70c8f6"
+		const secondSHA = "1f2e3e9245b159006cf2103becd51c5c1b6ec868"
```
xytan0056, line 458: can this be in Makefile envs instead?

integration/integration_test.go
```diff
@@ -0,0 +1,476 @@
...
+		switch item := msg.GetItem().(type) {
+		case *pb.GetChangedTargetsResponse_ChangedTargets:
+			if item.ChangedTargets != nil {
+				changedTargets = append(changedTargets, item.ChangedTargets.GetChangedTargets()...)
+			}
+		case *pb.GetChangedTargetsResponse_Metadata:
+			metadata = mergeMetadata(metadata, item.Metadata)
+		}
+	}
+	_ = stream.CloseSend()
```
xytan0056: let's defer this or make this a functional call. otherwise if `require.NoError` fails, this is a leak
```
func(){
  defer steam.Closesend()
  for {...}
}()
```

integration/integration_test.go
```diff
@@ -0,0 +1,476 @@
...
+		},
+	})
+	require.NoError(t, err, "failed to initiate GetTargetGraph stream")
+
+	raw := drainTargetGraphStream(t, stream)
+	require.NotNil(t, raw.metadata, "expected metadata in response")
+	require.NotEmpty(t, raw.metadata.GetTargetIdMapping())
+	require.NotEmpty(t, raw.metadata.GetRuleTypeMapping())
+
+	assert.Equal(t, 3673, len(raw.targets), "expected exact target count for pinned SHA")
```
xytan0056, line 374: this might be a bit fragile, cause if we upgrade bazel query with some options or change the hashing algorithm, the number can change. 
How about only check a few known targets and their dependency counts?

JacobOaks: Makes sense, happy to change it if we don't think it's valuable. I was thinking of it as a sort of "golden"-like assertion. As in, even if it gets modified frequently, I'm wondering if it's still useful to have as a heads up? Is there value in knowing/being "forced" to investigate (by means of this test failing) how graph size changes with updates we make to tango? What do you think?

xytan0056: it'll end up a [change detector](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) test. `pinnedSHA` is pinned, so people can just update that and blindly change whatever number asserted. Maybe a assert.Greater/Less range?

xytan0056:
@sywhang mentioned the integ test can be done in another approach. Could you take a look?

sywhang:
> @sywhang mentioned the integ test can be done in another approach. Could you take a look?

That's gonna take a while since we need to write various utils for emulating the git repo workload and do some fancy stuff.

Let's check in at least one E2E test that we can use, since we're about to do some big refactoring.

integration/integration_test.go
sywhang: followed up offline.

sywhang (approved):

integration/integration_test.go
JacobOaks: Not sure it makes much sense to do that. These test cases are asserting on expected changed targets for a specific set of SHAs. The expectations are hard-coded, so you can't really change the SHAs using env vars or the test will just.. fail. I think it makes more sense to keep the input next to the expected output.

I think a potential extension eventually would be to move all the shas & expected output into some sort of testdata files rather than hard-coding them, but this is already long-running as it is so not sure it matters until we can support running a lot of these tests somehow.

sywhang (approved):

xytan0056 (approved):

PR #158, opened 2026-07-07
docs(observability/metrics): document the metrics package

xytan0056:
add Tango Metrics Inventory RFC into the repo. Covers the Emitter contract and its usage in callsites

### DIFF
--- a/docs/observability/metrics.md
+++ b/docs/observability/metrics.md
@@ -7,6 +7,7 @@ Tango's metrics library. A thin, concrete wrapper over `tally.Scope` that pins t
 - op-as-subscope emission: an instrument bound as `(op, name)` lands under `<scope>.<op>.<name>`
 - a concrete, Tally-backed `Emitter` that binds instruments and copies tags before retaining them
 - explicit no-op construction for deployments that intentionally do not emit
+- a small set of shared vocabulary constants — operation names, tag keys, and result values — so every operation tags outcomes the same way
 
 It does not own metric names or histogram buckets from callers, such as controller, Bazel, storage, graphrunner etc. Those live in the package that implements each operation.
 
@@ -15,6 +16,7 @@ It does not own metric names or histogram buckets from callers, such as controll
 ```
 observability/metrics/
 ├── emitter.go       concrete Tally-backed emitter
+├── names.go         shared op / tag-key / result-value constants
 └── emitter_test.go
 
 controller/request_metrics.go   request lifecycle + buckets
@@ -31,30 +33,100 @@ graphrunner/metrics.go          graph runner metrics + buckets
 package metrics
 
 type Emitter struct {
-    scope tally.Scope
+    scope    tally.Scope
+    baseTags map[string]string
 }
 
-// New creates an emitter for scope. A nil scope is a configuration error.
-func New(scope tally.Scope) (*Emitter, error)
+// New creates an emitter for scope. A nil scope is a configuration error;
+// callers that intentionally emit nothing use Nop instead.
+func New(scope tally.Scope) (*Emitter, error) {
+    if scope == nil {
+        return nil, errors.New("metrics: nil scope")
+    }
+    return &Emitter{scope: scope}, nil
+}
 
-// Nop creates an explicitly disabled emitter.
-func Nop() *Emitter
+// Nop returns an explicitly disabled emitter backed by tally.NoopScope.
+func Nop() *Emitter { return &Emitter{scope: tally.NoopScope} }
 
-// Tagged returns an emitter with additional fixed tags. It copies tags and
-// does not modify its parent.
-func (e *Emitter) Tagged(tags map[string]string) *Emitter
+// Tagged returns a child that merges tags into every instrument it binds. It
+// copies tags and does not modify its parent.
+func (e *Emitter) Tagged(tags map[string]string) *Emitter {
+    if len(tags) == 0 {
+        return e
+    }
+    merged := make(map[string]string, len(e.baseTags)+len(tags))
+    for k, v := range e.baseTags {
+        merged[k] = v
+    }
+    for k, v := range tags {
+        merged[k] = v
+    }
+    return &Emitter{scope: e.scope, baseTags: merged}
+}
 
 // Counter binds a counter under <scope>.<op>.<name>.
-func (e *Emitter) Counter(op, name string) tally.Counter
+func (e *Emitter) Counter(op, name string) tally.Counter {
+    return e.opScope(op).Counter(name)
+}
 
 // DurationHistogram binds a duration histogram using owner-selected buckets.
-func (e *Emitter) DurationHistogram(op, name string, buckets tally.DurationBuckets) tally.Histogram
+func (e *Emitter) DurationHistogram(op, name string, b tally.DurationBuckets) tally.Histogram {
+    return e.opScope(op).Histogram(name, b)
+}
 
 // ValueHistogram binds a value histogram using owner-selected buckets.
-func (e *Emitter) ValueHistogram(op, name string, buckets tally.ValueBuckets) tally.Histogram
+func (e *Emitter) ValueHistogram(op, name string, b tally.ValueBuckets) tally.Histogram {
+    return e.opScope(op).Histogram(name, b)
+}
+
+// opScope applies the op subscope and any accumulated tags.
+func (e *Emitter) opScope(op string) tally.Scope {
+    s := e.scope.SubScope(op)
+    if len(e.baseTags) > 0 {
+        s = s.Tagged(e.baseTags)
+    }
+    return s
+}
 ```
 
 Only `New` returns an error, and only to surface a nil scope — a wiring mistake, not a runtime condition. The binding methods take constant `op`/`name` arguments and have no reachable failure path, so they return the instrument directly. Instruments are usually bound once during component construction and stored on a local metrics struct; a component whose metrics carry a per-request tag (see [Request metrics](#request-metrics)) builds its struct per request instead.
+
+## Conventions
+
+Metric names and buckets are domain-local, but the *outcome vocabulary* is shared: an operation that tags `result=success` and one that tags `result=succeeded` will not aggregate together, and a dashboard can't discover that mismatch. So operation names, tag keys, and result values live in `metrics/names.go` and every operation draws from them.
+
+```go
+package metrics
+
+// Operation names — the <op> subscope for Tango's core operations. Extension
+// operations declare their own const next to the emit site.
+const (
+    OpGetTargetGraph        = "get_target_graph"
+    OpGetChangedTargets     = "get_changed_targets"
+    OpGetChangedTargetGraph = "get_changed_target_graph"
+    OpGetGraph              = "get_graph"
+    OpCompareTargetGraphs   = "compare_target_graphs"
+    OpGraphRunner           = "graph_runner"
+)
+
+// Tag keys.
+const (
+    TagRepo      = "repo"
+    TagResult    = "result"
+    TagOperation = "operation"
+)
+
+// Result values for TagResult.
+const (
+    ResultSuccess = "success"
+    ResultFailure = "failure"
+    ResultHit     = "hit"
+    ResultMiss    = "miss"
+)
+```
+
+An outcome is recorded as one latency histogram tagged `result=success|failure` — not a pair of `succeeded`/`failed` counters. A histogram already carries a per-bucket count, so the success and failure *rates* are derived from it (sum the buckets, group by `result`), and a separate outcome counter would be redundant. See [Request metrics](#request-metrics).
 
 ## Dependency direction
 
@@ -68,7 +140,7 @@ infrastructure       -> local metrics      -> observability/metrics -> tally
 
 ## Domain-local metrics
 
-A domain component defines a small metrics struct in its own vocabulary. When its metrics carry no per-request tag, it binds them — and its bucket policy — once, at construction. Buckets are a default owned by the component, overrideable. For example,
+A domain component defines a small metrics struct in its own vocabulary. Instruments whose tags are fixed for the component's lifetime are bound once, at construction; an instrument whose tags depend on the outcome (the `result`-tagged latency histogram) is bound at record time — tally caches it by name+tags, and the `result` value is bounded (`success`/`failure`), so this stays low-cardinality. Buckets are a default owned by the component, overrideable. For example,
 
 ```go
 package bazel
@@ -82,9 +154,8 @@ var (
 )
 
 type queryMetrics struct {
-    succeeded   tally.Counter
-    failed      tally.Counter
-    duration    tally.Histogram
+    emitter     *metrics.Emitter
+    buckets     tally.DurationBuckets
     targetCount tally.Histogram
 }
 
@@ -96,10 +167,24 @@ func newQueryMetrics(e *metrics.Emitter, dur tally.DurationBuckets, tgt tally.Va
         tgt = defaultQueryTargetBuckets
     }
     return &queryMetrics{
-        succeeded:   e.Counter(opQuery, "succeeded"),
-        failed:      e.Counter(opQuery, "failed"),
-        duration:    e.DurationHistogram(opQuery, "duration", dur),
+        emitter:     e,
+        buckets:     dur,
         targetCount: e.ValueHistogram(opQuery, "target_count", tgt),
+    }
+}
+
+// record emits query latency tagged with the outcome (success/failure counts
+// are derived from this histogram) and the target count on success.
+func (m *queryMetrics) record(start time.Time, targets int, err error) {
+    result := metrics.ResultSuccess
+    if err != nil {
+        result = metrics.ResultFailure
+    }
+    m.emitter.Tagged(map[string]string{metrics.TagResult: result}).
+        DurationHistogram(opQuery, "duration", m.buckets).
+        RecordDuration(time.Since(start))
+    if err == nil {
+        m.targetCount.RecordValue(float64(targets))
     }
 }
 ```
@@ -108,46 +193,54 @@ The bazel client's exported `Params` carries the `Emitter` and optional bucket o
 
 ## Request metrics
 
-Request outcome policy lives at the controller boundary. The pattern is controller-local, not part of the emitter package. Outcomes are broken down by `repo`, so a `requestMetrics` is built per request from a repo-tagged emitter; tally caches instruments by name+tags, so each `(op, repo)` series is bound once (lazily) and reused. Duration buckets follow the same default-and-override rule as domain-local metrics.
+Request outcome policy lives at the controller boundary — the pattern is controller-local, not part of the emitter package. A request emits two things: a `started` counter when it begins, and a single latency histogram tagged with the outcome when it completes. The success/failure *count* is derived from the histogram (sum its buckets, group by `result`), so there is no separate outcome counter. Outcomes are broken down by `repo`, so `requestMetrics` is built per request from a repo-tagged emitter; tally caches by name+tags, so each `(op, repo)` series is bound once and reused. Duration buckets follow the same default-and-override rule as domain-local metrics.
 
 ```go
 // default request-latency buckets
 var defaultRequestDurationBuckets = tally.MustMakeExponentialDurationBuckets(time.Millisecond, 3, 10) // 1ms .. ~59s
 
 type requestMetrics struct {
-    succeeded tally.Counter
-    failed    tally.Counter
-    duration  tally.Histogram
+    emitter *metrics.Emitter
+    op      string
+    buckets tally.DurationBuckets
+    started tally.Counter
 }
 
 func newRequestMetrics(e *metrics.Emitter, op string, buckets tally.DurationBuckets) *requestMetrics {
     if len(buckets) == 0 {
         buckets = defaultRequestDurationBuckets
     }
-    return &requestMetrics{
-        succeeded: e.Counter(op, "succeeded"),
-        failed:    e.Counter(op, "failed"),
-        duration:  e.DurationHistogram(op, "duration", buckets),
-    }
+    return &requestMetrics{emitter: e, op: op, buckets: buckets, started: e.Counter(op, "started")}
 }
 
-// Record records the request's duration and outcome. Total request rate is
-// succeeded + failed.
-func (m *requestMetrics) Record(start time.Time, err error) {
-    m.duration.RecordDuration(time.Since(start))
-    if err == nil {
-        m.succeeded.Inc(1)
-        return
+// requestAttempt is one in-progress request.
+type requestAttempt struct {
+    owner *requestMetrics
+    start time.Time
+}
+
+// Begin records that the request started and returns the attempt to Complete.
+func (m *requestMetrics) Begin() *requestAttempt {
+    m.started.Inc(1)
+    return &requestAttempt{owner: m, start: time.Now()}
+}
+
+// Complete records latency into a single histogram tagged with the outcome;
+// success/failure counts are derived from it.
+func (a *requestAttempt) Complete(err error) {
+    result := metrics.ResultSuccess
+    if err != nil {
+        result = metrics.ResultFailure
     }
-    m.failed.Inc(1)
+    a.owner.emitter.Tagged(map[string]string{metrics.TagResult: result}).
+        DurationHistogram(a.owner.op, "latency", a.owner.buckets).
+        RecordDuration(time.Since(a.start))
 }
 ```
 
-The controller keeps the emitter and bucket policy, and builds `requestMetrics` per request off a repo-tagged emitter:
+The handler begins the attempt off a repo-tagged emitter and completes it in a defer:
 
 ```go
-const tagRepo = "repo"
-
 func newController(p Params) *controller {
     return &controller{
         emitter:                p.Emitter,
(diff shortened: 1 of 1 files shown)


docs/observability/metrics.md
```diff
@@ -0,0 +1,112 @@
...
+
+Tango is a library. Consumers wire it into their own `fx` graph with their own `tally.Scope`, and often deploy multiple flavors (long-running server, batch job, one-shot CLI) side-by-side. Without a shared convention every deployment invents its own metric names and tag layout, and dashboards devolve into per-name merges.
+
+This package owns:
+
+- op-as-subscope emission (`e.Inc("get_target_graph", "requests")` lands under `<scope>.get_target_graph.requests`)
+- shared metric-name, op-name, and tag-key constants
+- default histogram buckets sized for RPC latency and large target counts
+- an in-flight gauge helper that owns the atomic counter callers otherwise reinvent
+- a failure classifier that reuses the errors package
```
xytan0056: @justinwon777 place holder for errors package design RFC

docs/observability/metrics.md
```diff
@@ -0,0 +1,137 @@
...
+├── options.go   — Option, WithTags, WithDurationBuckets, WithValueBuckets
+├── buckets.go   — default histogram buckets
+├── names.go     — Op*, metric name, Tag*, Result* constants
+├── failure.go   — RecordFailure
+└── context.go   — WithEmitter, FromContext
+```
+
+## Emitter
+
+`Emitter` is an interface. `New(scope)` returns the default tally-backed implementation; `New(nil)` returns one backed by `tally.NoopScope` so callers do not need to special-case unwired metrics. This also serves as a noop fallback.
```
sywhang:
```suggestion
`Emitter` is an interface. `New(scope)` returns the default tally-backed implementation; `New(nil)` returns one backed by `tally.NoopScope` so callers do not need to special-case metrics using alternative metrics library than tally. This also serves as a no-op fallback which could be useful for things like tests.
```

docs/observability/metrics.md
```diff
@@ -0,0 +1,137 @@
...
+  durationBuckets tally.DurationBuckets
+  valueBuckets    tally.ValueBuckets
+}
+```
+
+`New(scope)` returns the default tally-backed implementation; `New(nil)` returns one backed by `tally.NoopScope` so callers do not need to special-case unwired metrics.
+
+`Tagged` returns a child Emitter that adds tags to every subsequent emission. The parent is unchanged. The child shares the in-flight counter map, so a `TrackInFlight` increment on a parent and its decrement on a tagged child balance to zero.
+
+Options are variadic and compose. `WithTags` merges into the base tag set with later-wins semantics on key collision. `WithDurationBuckets` / `WithValueBuckets` override the emitter's default histogram buckets on a single call.
```
sywhang: can this section be written as a docstring with func signature?

```go

// New returns the default Tally-backed Emitter implementation.
func New(scope tally.Scope) Emitter 

// and so on... 
```

docs/observability/metrics.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Op-as-subscope
```
sywhang: what does this mean??

docs/observability/metrics.md
```diff
@@ -0,0 +1,137 @@
...
+
+Every emit call takes an op name as its first argument. `e.Inc("get_target_graph", "requests")` emits under `<scope>.get_target_graph.requests`. Consumers that own their own ops (an extension RPC, a background job) should declare `const opXYZ = "xyz"` next to the emit site rather than adding to `names.go`.
+
+## TrackInFlight
```
sywhang: same thing here, let's make this into an API form.

docs/observability/metrics.md
```diff
@@ -0,0 +1,137 @@
...
+defer c.emitter.TrackInFlight(metrics.OpGetTargetGraph)()
+```
+
+The counter map lives on the root Emitter, so a `Tagged` child shares the counter with its parent — a paired increment/decrement across a `Tagged` boundary always refers to the same gauge series. The emitted gauge carries `TagOperation=op` so a single time-series carries the count for every operation, split by op in the dashboard.
+
+## RecordFailure
+
+Every RPC calls one of:
+
+- `e.Inc(op, Requests, WithTags({result: success}))` on the happy path
```
sywhang: `Inc` and `RecordFailure` aren't intuitive names for counterparts of each other. can we rename?

sywhang: Also, not sure why is `Inc` and `failure`/`success` is a thing. Do we need this?

docs/observability/metrics.md
```diff
@@ -0,0 +1,137 @@
...
+   if !ok { return noopEmitter }
+   return e
+}
+```
+
+The RPC handler applies request-scope tags once and installs the tagged Emitter on the context. Helpers pull it back out with `FromContext`, which never returns nil — a missing key falls back to a package-level noop.
+
+## Constants
+
+All shared constants live in `names.go`: operation names (`OpGetTargetGraph`, `OpGetChangedTargets`, ...), metric names (`Requests`, `TotalDuration`, `InFlightRequests`, ...), tag keys (`TagRepo`, `TagEmitter`, `TagResult`, `TagFailureType`, `TagFailureSource`, ...), and `Result*` values (`ResultSuccess`, `ResultFail`, `ResultHit`, `ResultMiss`).
```
sywhang: where they live isn't relevant to this RFC.

Can we make this into a table form?

docs/observability/metrics.md
xytan0056: Op-as-subscope means passed op will be a subscope,  called like 
```
func (e *defaultEmitter) Inc(op, name string, opts ...Option) {
	...
	e.Subscope(op).Counter(name).Inc(1)
}
```
so the metric name is `controller.get_target_graph.TreehashCacheLookup`

docs/observability/metrics.md
xytan0056: good point. added symmetrical helper and documented generic purpose of `Inc`

docs/observability/metrics.md
xytan0056: good point

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
+# observability/metrics
+
+Tango's metrics library. A thin, opinionated wrapper over `tally.Scope` that pins the naming, and tag conventions Tango consumers emit so dashboards can query by tag rather than by name merging.
+
+## Why
+
+Tango is a library. Consumers wire it into their own `fx` graph with their own `tally.Scope`, and often deploy multiple flavors (long-running server, batch job, one-shot CLI) side-by-side. Without a shared convention every deployment invents its own metric names and tag layout, and dashboards devolve into per-name merges.
+
+This package owns:
```
sbalabanov: should this file be README.md under `observability/metrics` then?

docs/observability/metrics.md
sbalabanov: why metrics should own a classifier?

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+
+- op-as-subscope emission (`e.Inc("get_target_graph", "requests")` lands under `<scope>.get_target_graph.requests`)
+- shared metric-name, op-name, and tag-key constants
+- default histogram buckets sized for RPC latency and large target counts
+- an in-flight gauge helper that owns the atomic counter callers otherwise reinvent
+- a failure classifier that reuses the errors package
+- context-based emitter propagation so downstream helpers do not need the emitter threaded through every signature
```
sbalabanov: they still need to thread context object though.
For the record, I do not necessarily agree that this is better approach than explicit signature-based, but not going to make it blocking.

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+├── options.go   — Option, WithTags, WithDurationBuckets, WithValueBuckets
+├── buckets.go   — default histogram buckets
+├── names.go     — Op*, metric name, Tag*, Result* constants
+├── failure.go   — RecordFailure
+└── context.go   — WithEmitter, FromContext
+```
+
+## Emitter
+
+`Emitter` is an interface. `New(scope)` returns the default tally-backed implementation; `New(nil)` returns one backed by `tally.NoopScope`, so callers do not need to special-case metrics using alternative metrics library than tally. This also serves as a no-op fallback which could be useful for tests.
```
sbalabanov: it is unlikely that clients could use anything other than Tally; tally is used for all of the internal functions that clients won't override.
The extension point is typically inverted: clients use Tally library but can provide a custom backend.

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+
+```go
+type Emitter interface {
+  Inc(op, name string, opts ...Option)
+  Gauge(op, name string, v float64, opts ...Option)
+  RecordDur(op, name string, d time.Duration, opts ...Option)
+  RecordCount(op, name string, v int64, opts ...Option)
```
sbalabanov: what is a difference between this and `Inc` ?

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+
+## Emitter
+
+`Emitter` is an interface. `New(scope)` returns the default tally-backed implementation; `New(nil)` returns one backed by `tally.NoopScope`, so callers do not need to special-case metrics using alternative metrics library than tally. This also serves as a no-op fallback which could be useful for tests.
+
+```go
+type Emitter interface {
+  Inc(op, name string, opts ...Option)
+  Gauge(op, name string, v float64, opts ...Option)
+  RecordDur(op, name string, d time.Duration, opts ...Option)
```
sbalabanov: with Histogram approach it is typical to record both duration and outcome (i.e. success or failure) into the same series of low-cardinality metrics.
How to do it with the proposal?
Something like
`ctx.getEmitter().Tagged("outcome": "success").RecordDur(what_is_op, and_what_is_name, duration.Second(1))` ?

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+
+These are helper functions wrapping `Inc` calls for top-level request recording.
+
+```go
+// RecordSuccess emits a requests counter tagged with result=success.
+func RecordSuccess(e Emitter, op string)
+
+// RecordFailure emits a requests counter tagged with result=fail plus
+// failure information derived from err. No-op on nil.
+func RecordFailure(e Emitter, op string, err error)
```
sbalabanov: need a cross reference to error classification.
what is "additional failure information"?

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+## Op-as-subscope
+
+Every emit call takes an op name as its first argument. For example, `e.Inc("get_target_graph", "requests")` emits under `<scope>.get_target_graph.requests`. Consumers that own their own ops (an extension RPC, a background job) should declare `const opXYZ = "xyz"` next to the emit site rather than adding to `names.go`.
+
+## RecordSuccess / RecordFailure
+
+These are helper functions wrapping `Inc` calls for top-level request recording.
+
+```go
+// RecordSuccess emits a requests counter tagged with result=success.
```
sbalabanov: is the number of reported error types is limited, it is more efficient for the metrics backend to store the outcome in the same tag, i.e. result=success|infra_error|infra_retryable_error|user_error

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+`Inc` is the generic counter for everything else — cache lookups, patches applied, retries, workspace leases:
+
+```go
+e.Inc(op, TreehashCacheLookup, WithTags(map[string]string{TagResult: string(ResultHit)}))
+e.Inc(op, Patch...)
+```
+
+## TrackInFlight
+
+Gauges are update-only, so an in-flight counter needs an owning atomic somewhere. Rather than force every call site to allocate its own `atomic.Int64` and remember to update the gauge, the Emitter.TrackInFlight owns a per-op `*int64`:
```
sbalabanov: what is the practical example where gauges need to be used for Tango?
Gauge is a pretty nasty metric type when it comes to distributed systems with multiple running instances, and should be in general avoided.

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+```go
+type emitterKey struct{}
+var noopEmitter = New(nil)
+
+func WithEmitter(ctx context.Context, e *Emitter) context.Context {
+   return context.WithValue(ctx, emitterKey{}, e)
+}
+func FromContext(ctx context.Context) *Emitter { // should never return nil
+   e, ok := ctx.Value(emitterKey{}).(*Emitter)
+   if !ok { return noopEmitter }
```
sbalabanov: panic or `error` ? :)

sbalabanov: at least, let RFC discuss tradeoffs

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+    })
+    return &controller{emitter: emitter, /* ... */}
+}
+
+func (c *controller) GetChangedTargets(req *pb.GetChangedTargetsRequest, stream pb...) (retErr error) {
+    defer c.emitter.TrackInFlight(metrics.OpGetChangedTargets)()
+    e := c.emitter.Tagged(map[string]string{
+        metrics.TagRepo: common.ToShortRemote(req.GetFirstRevision().GetRemote()),
+    })
+    ctx := metrics.WithEmitter(stream.Context(), e)
```
sbalabanov: should combine app and stream contexts

sbalabanov (changes requested):
Missing pieces:
- error classification integration
- working with histograms for both status and duration and an example on custom buckets

docs/observability/metrics.md
```diff
@@ -0,0 +1,223 @@
...
+    start := time.Now()
+    defer func() {
+        e.RecordDur(metrics.OpGetChangedTargets, metrics.TotalDuration, time.Since(start))
+        if retErr != nil {
```
sbalabanov: this could be a helpre function as well to avoid writing the same if statement in each `defer`.
In fact, with histograms it would be a very single helper function call accepting duration and error objects (error could be nil).

docs/observability/metrics.md
xytan0056: RecordFailure needs to call classifier to get errorCode to count reasons for failure. The err needs to be passed in. But fold this in RecordRequest as suggested below. 
good point, will add reference to err classifier

docs/observability/metrics.md
xytan0056: yep that's the intention, `RecordSuccess` `RecordFailure` will both use `result=` tag. 
I folded these in `RecordRequest` as suggested below

docs/observability/metrics.md
xytan0056: Considering how heavy graph computation could be, I think it's beneficial to track inflight requests/leased workspaces per instance. Gauge is the natural fit here, this way users can have visibility into how much scale the service needs (e.g. how many workers exhausted workspace pool). Especially for native_orchestrator, where the computation happen on hosts.

docs/observability/metrics.md
xytan0056: sure I'll add some explanations.
imo, 
- observability is not busisness logic, not having metrics emission should not crash the server or fail the request.
- in practice, a missing emitter means someone forgot to call `WithEmitter` in the handler -> lose metrics -> notice in dashboards -> fix the settitng, it
- other wise every callsite has to handle this error, `if err := FromContext(ctx); err != nil {}`

docs/observability/metrics.md
xytan0056: Good point. Yes, if a function needs to send metrics, it should explictly extract from context, but that should be the standard idiom (context has to be there, no context is worse that no metrics). I'll add this bit in the doc
This is the same for all other request scoped things like logger, tracings, signals etc. Emitter is what all downstream functions need but not required for business logic. 
 Signature based will just require the arg in every callsite -- like passing scope into hashing even though it's not needed in the actual logic, plus we'll have to create a subscope and be careful not to mix up with the one from parent

docs/observability/metrics.md
xytan0056: just discussed that we will just emit classified error code instead. Will remove

docs/observability/metrics.md
xytan0056:
```
emitter:= FromContext(ctx)
emitter.RecordDur(op, name, duration.Second(1), WithTags(map[string]string{"result", "success"}))
```

docs/observability/metrics.md
xytan0056: done

docs/observability/metrics.md
xytan0056: following https://github.com/uber-go/fx/tree/d5da5b04ac906bfbad8b400baeee9b970c1be6f3/docs/src
this is meant to be usage doc

docs/observability/metrics.md
xytan0056: good point. done

docs/observability/metrics.md
xytan0056: `Inc` is counter for counting calls, `RecordCount` is histogram for recording # of targets. Added explanation to make it clear

docs/observability/metrics.md
xytan0056: good point, reworded

docs/observability/metrics.md
sbalabanov: In practice, when you track per-instance metrics, cardinality starts to become a problem.
It is almost always better to report two counter types of metrics instead: first when the request started, and second when the request finished (as a histogram, with latency and an outcome).
Rely on logging/log aggregation for high-cardinality tracking/debugging.

docs/observability/metrics.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Emitter
+
+`Emitter` is an interface. `New(scope)` returns the default tally-backed implementation; `New(nil)` returns a no-op, useful for tests.
+
+```go
+type Emitter interface {
+  Inc(op, name string, opts ...Option)          // counter: +1 each call
+  Gauge(op, name string, v float64, opts ...Option)
+  RecordDur(op, name string, d time.Duration, opts ...Option)  // duration histogram
```
sbalabanov: if count is histogram than what is RecordDur?
I think the abstraction should be contracted more explicitly

xytan0056:
updated based on discussion offline.

docs/observability/metrics.md
```diff
@@ -0,0 +1,196 @@
...
+        requestDurationBuckets: p.RequestDurationBuckets,
+    }
+}
+
+func (c *controller) GetChangedTargets(req *pb.GetChangedTargetsRequest, stream pb.Tango_GetChangedTargetsServer) (retErr error) {
+    // repo is bounded (allow-listed) and normalized to a stable key.
+    e := c.emitter.Tagged(map[string]string{
+        tagRepo: common.ToShortRemote(req.GetFirstRevision().GetRemote()),
+    })
+    m := newRequestMetrics(e, opGetChangedTargets, c.requestDurationBuckets)
```
xytan0056: requestmetrics need to be per-repo, so this is initialized per request.

docs/observability/metrics.md
```diff
@@ -0,0 +1,196 @@
...
+    return &requestMetrics{
+        succeeded: e.Counter(op, "succeeded"),
+        failed:    e.Counter(op, "failed"),
+        duration:  e.DurationHistogram(op, "duration", buckets),
+    }
+}
+
+// Record records the request's duration and outcome. Total request rate is
+// succeeded + failed.
+func (m *requestMetrics) Record(start time.Time, err error) {
```
xytan0056: dropped the `begin/complete` pattern, because total called is just succeeded + failed. After we decide to drop gauge, there's nothing to inc/dec at the begining anymore. All the request metrics can be done in a single defer
```
start := time.Now()
defer func() { m.Record(start, retErr) }()
```

docs/observability/metrics.md
```diff
@@ -0,0 +1,196 @@
...
+package metrics
+
+type Emitter struct {
+    scope tally.Scope
+}
+
+// New creates an emitter for scope. A nil scope is a configuration error.
+func New(scope tally.Scope) (*Emitter, error)
+
+// Nop creates an explicitly disabled emitter.
```
sbalabanov: how would it do so if Emitter is not an interface? With noop tally scope?

docs/observability/metrics.md
```diff
@@ -0,0 +1,196 @@
...
+func Nop() *Emitter
+
+// Tagged returns an emitter with additional fixed tags. It copies tags and
+// does not modify its parent.
+func (e *Emitter) Tagged(tags map[string]string) *Emitter
+
+// Counter binds a counter under <scope>.<op>.<name>.
```
sbalabanov: A real implementation should have better description

docs/observability/metrics.md
```diff
@@ -0,0 +1,196 @@
...
+
+func newQueryMetrics(e *metrics.Emitter, dur tally.DurationBuckets, tgt tally.ValueBuckets) *queryMetrics {
+    if len(dur) == 0 {
+        dur = defaultQueryDurationBuckets
+    }
+    if len(tgt) == 0 {
+        tgt = defaultQueryTargetBuckets
+    }
+    return &queryMetrics{
+        succeeded:   e.Counter(opQuery, "succeeded"),
```
sbalabanov: Does it mean that some another metric, like hashing metric, will also hardcode the action? In which case "query" metric will look like `name=get_targets.query.succeeded` and "hashing" metric might be different as `name=get_targets.query.success` or `name=get_targets.query.finished`
Should we standardize tag naming?

docs/observability/metrics.md
```diff
@@ -0,0 +1,196 @@
...
+        duration:  e.DurationHistogram(op, "duration", buckets),
+    }
+}
+
+// Record records the request's duration and outcome. Total request rate is
+// succeeded + failed.
+func (m *requestMetrics) Record(start time.Time, err error) {
+    m.duration.RecordDuration(time.Since(start))
+    if err == nil {
+        m.succeeded.Inc(1)
```
sbalabanov: no need for a separate counter for the outcome.
They could be recorded into the same histogram as duration under a separate tag.
The query would do "sum" function over a histogram time-series to understand the number of failed or successful requests.
Histogram is itself just a glorified counter.

docs/observability/metrics.md
sbalabanov: Begin may want to record that the operation started

docs/observability/metrics.md
sbalabanov: Is it really needed to create a `metrics` structure per each metric type emitted?
SubmitQueue uses simpler structure with standalone helper functions and a general scope object:
https://github.com/uber/submitqueue/blob/main/submitqueue/orchestrator/controller/start/start.go#L76-L77
https://github.com/uber/submitqueue/blob/main/submitqueue/orchestrator/controller/validate/validate.go#L92-L93

sbalabanov (changes requested):

docs/observability/metrics.md
xytan0056: Yes 
```
func Nop() *Emitter { return &Emitter{scope: tally.NoopScope} }
```

docs/observability/metrics.md
xytan0056: good point. standardized into constants

docs/observability/metrics.md
xytan0056: there's a concern of scopes being tagged everywhere which is expensive.
I updated with a memoization approach in each domain, what do you think?

docs/observability/metrics.md
```diff
@@ -0,0 +1,348 @@
...
+
+func (e *Emitter) DurationHistogram(op, name string, b tally.DurationBuckets) tally.Histogram {
+    return e.scope.SubScope(op).Histogram(name, b)
+}
+
+func (e *Emitter) ValueHistogram(op, name string, b tally.ValueBuckets) tally.Histogram {
+    return e.scope.SubScope(op).Histogram(name, b)
+}
+
+func TagKey(tags map[string]string) string {
```
sbalabanov: this is why?

docs/observability/metrics.md
```diff
@@ -0,0 +1,348 @@
...
+        b.WriteString(k)
+        b.WriteByte(':')
+        b.WriteString(tags[k])
+    }
+    return b.String()
+}
+```
+
+
+Every instrumented operation emits exactly two metrics under the `start`/`finish` convention:
```
sbalabanov: "exactly two"
well I think it can emit more (like number of changed targets) if it wants to, so should say "at least two"?

docs/observability/metrics.md
```diff
@@ -0,0 +1,348 @@
...
+
+| Metric | Kind | Records |
+|---|---|---|
+| `<scope>.<op>.start` | counter | operations attempted |
+| `<scope>.<op>.finish` | histogram, `result`-tagged | operations completed — latency + outcome |
```
sbalabanov: result and outcome same thing

docs/observability/metrics.md
```diff
@@ -0,0 +1,348 @@
...
+    ResultMiss      = "miss"
+)
+```
+```go
+// Outcome maps an error to a result tag value.
+func Outcome(err error) string {
+    switch {
+    case err == nil:
+        return ResultSuccess
+    case errors.Is(err, context.Canceled), errors.Is(err, context.DeadlineExceeded):
```
sbalabanov: deadlineexceeded is not cancelled, it is a legitimate timeout error, in most cases infra or infra_retryable

xytan0056: makes sense. I was thinking if the the context is set at client side and exceeds deadline, then client should set a higher timeout. But got your point that tango is probablly too slow to meet expectation in this case

docs/observability/metrics.md
xytan0056: synced offline. Will go with helper approach

sbalabanov (approved):

PR #161, opened 2026-07-08
docs: Tango Errors RFC

justinwon777:
Adds a Tango Errors RFC covering the `core/errors` package and its APIs


docs/errors/errors.md
```diff
@@ -0,0 +1,241 @@
...
+func (code ErrorCode) String() string
+```
+
+String returns the string form of the code: `"bad_request"`, `"internal"`, `"internal_retryable"`, or `"unknown"` as the default.
+
+### type FailureSource
+
+```go
+type FailureSource interface {
+	source() // unexported method to prevent outsiders from implementing
```
sywhang: Impl detail, remove

docs/errors/errors.md
```diff
@@ -0,0 +1,241 @@
...
+}
+```
+
+`FailureSource` represents the component where an error occurred. This tag provides extra granularity for failures so that if errors spike, it's easy to know which part of Tango they're coming from. The interface is implemented with an unexported method, so outside packages cannot create their own sources and pollute the metric's cardinality. All valid sources are defined in this package.
+
+#### Failure Source values
+
+```go
+var (
+	FailureSourceUnknown       FailureSource
```
sywhang: Can we document these

docs/errors/errors.md
```diff
@@ -0,0 +1,241 @@
...
+
+```go
+type TangoError struct {
+	failureSource FailureSource
+	err           error
+	errorCode     ErrorCode
+}
+```
+
+`TangoError` holds all the information Tango needs to classify and return a proto error: the underlying error, the `FailureSource` it originated from, and its `ErrorCode`.
```
sywhang: Add a bit more detail here on what this is for

docs/errors/errors.md
```diff
@@ -0,0 +1,241 @@
...
+
+The three constructors below build a `TangoError` from an error and a `FailureSource`, classifying it with a fixed `ErrorCode`. In all three, if `err` already wraps a `TangoError`, the returned error preserves the original's source and code instead of the ones passed in.
+
+#### func NewInternal
+
+```go
+func NewInternal(src FailureSource, err error) error
+```
+
+NewInternal wraps err as a `TangoError` classified `ERROR_INTERNAL`.
```
sywhang: Need more description here. What's an internal error

docs/errors/errors.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### func NewUser
+
+```go
+func NewUser(src FailureSource, err error) error
+```
+
+NewUser wraps err as a `TangoError` classified `ERROR_BAD_REQUEST`.
```
sywhang: Same thing here. What is a user error

docs/errors/errors.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### func (\*TangoError) FailureSource
+
+```go
+func (te *TangoError) FailureSource() FailureSource
+```
+
+FailureSource returns the component that produced the error.
+
+## Package `mapper`
```
sywhang: Technically this is outside the scope of the error pkg. let's remove

docs/errors/errors.md
justinwon777: Is the error code definition at the top not enough? Or should I reiterate the definitions here?

sywhang: synced offline.

docs/errors/errors.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Proto
+
+`tango/proto/tango.proto` defines the wire-level error type returned by every RPC when it fails.
+
+```protobuf
+enum ErrorCode {
+    ERROR_UNKNOWN = 0;
+    ERROR_BAD_REQUEST = 1;
+    ERROR_INTERNAL = 2;
```
sbalabanov: we use "infra" and "user" terminology across the stack, I suggest this to be the same

docs/errors/errors.md
```diff
@@ -0,0 +1,202 @@
...
+message TangoError {
+    ErrorCode code = 1;
+    String message = 2;
+}
+```
+
+| Error Code | Definition |
+|---|---|
+| `ERROR_UNKNOWN` | Unexpected errors |
+| `ERROR_BAD_REQUEST` | Request validation errors (missing arg, invalid arg), canceled requests, user failures in builds |
```
sbalabanov: we should be able to differentiate between missing arg (which is a programming error on a caller side) vs build failure (which is a user error not related to the ecosystem correctness)

docs/errors/errors.md
```diff
@@ -0,0 +1,202 @@
...
+
+message TangoError {
+    ErrorCode code = 1;
+    String message = 2;
+}
+```
+
+| Error Code | Definition |
+|---|---|
+| `ERROR_UNKNOWN` | Unexpected errors |
```
sbalabanov: what is a difference between UNKNOWN and INTERNAL?

docs/errors/errors.md
```diff
@@ -0,0 +1,202 @@
...
+
+### type FailureSource
+
+```go
+type FailureSource interface {
+	String() string
+}
+```
+
+`FailureSource` represents the component where an error occurred. This tag provides extra granularity for failures so that if errors spike, it's easy to know which part of Tango they're coming from. The interface is implemented with an unexported method, so outside packages cannot create their own sources and pollute the metric's cardinality. All valid sources are defined in this package.
```
sbalabanov: in the example above, `String()` is exported

docs/errors/errors.md
```diff
@@ -0,0 +1,202 @@
...
+	FailureSourceRepoManager FailureSource = source("repomanager")
+)
+```
+
+### type TangoError
+
+```go
+type TangoError struct {
+	failureSource FailureSource
+	underlying    error
```
sbalabanov: better call it `err`

docs/errors/errors.md
justinwon777: updated the error code definitions and made the error codes in the constructors link back to the definition.

sbalabanov (changes requested):
Few things are missing in the RFC design:
- Clear error domain definition. What types of errors could ever be in the system? (see comments on data structures).
- Guidelines on using classifiers; where exactly should classification be done and what data will it use. In the provided example, bazel query error may be both infra (i.e. no bazel binary on the machine) or user (i.e. change is incompatible with the trunk). Infra might be retryable (i.e. bazilisk failed with http error) or not (i.e. incorrect argument list to this  specific bazel version).
- How internal classification translates to the API layer (GRPC codes and rich error object). GRPC defines its own codes which cannot be changed or extended, typically observable by common infrastructure (i.e. muttley).
- Missing dependency flag. We would like to compute inclusive and exclusive reliability metric for the service, i.e. it should differentiate between its own errors (tango process + subprocesses it runs) and external systems it calls (i.e. CI or storage).

docs/errors/errors.md
justinwon777: I was thinking our internal callers can have a mapping to "user" and "infra" if they want to use those terms, but for OSS "bad request" and "internal" are more commonly used, so it'd be easier for people to understand.

docs/errors/errors.md
justinwon777: `UNKNOWN` is used as a default if an error is never classified. We have a lot of places currently returning plain errors, so they will show up as unknown, and we can go through them and classify them. Sort of like how we use `generic` in our buildkite classifier.

docs/errors/errors.md
justinwon777: There's an additional unexported method not shown here.

docs/errors/errors.md
justinwon777: I was told by @sywhang that the convention is to call wrapped errs `underlying`

docs/errors/errors.md
justinwon777: I agree we should eventually have different error codes for different types of user failures like canceled and build failure, but I think we should take advantage of that once our error classification is more mature. For instance, right now there is no way for Tango to know why a buildkite build failed, so there is no use of having a build failure error code.

justinwon777:
@sbalabanov partial response, still looking into your GRPC codes comment

> * Guidelines on using classifiers; where exactly should classification be done and what data will it use. In the provided example, bazel query error may be both infra (i.e. no bazel binary on the machine) or user (i.e. change is incompatible with the trunk). Infra might be retryable (i.e. bazilisk failed with http error) or not (i.e. incorrect argument list to this  specific bazel version).

I added a more detailed example of the usage. This package's purpose is to provide the framework for callers to classify an error, and the caller is responsible for determining how to classify an error. At the controller layer, the mapper will handle all errors and convert them to the proto type (unclassified will be unknown error code). Some packages like bazel will need more work done to differentiate user and infra failures.

> * Missing dependency flag. We would like to compute inclusive and exclusive reliability metric for the service, i.e. it should differentiate between its own errors (tango process + subprocesses it runs) and external systems it calls (i.e. CI or storage).

Yes we could potentially add two more error codes ErrorInternalDependency and ErrorInternalDependencyRetryable. In a future task we can identify external systems that can use these and then add the error codes.

docs/errors/errors.md
sywhang: it's a wrapped error, also not an exported field so name here doesn't really matter

docs/errors/errors.md
sywhang: since this is a contract with rest of the stack like SQ, it's better to keep things consistent esp. if protos are all imported together - consider changing this to infra.

justinwon777: Ok, I don't have strong feelings about it, so I'd prefer doing what's better for our users in the stack. Changed to "user" and "infra".

docs/errors/errors.md
justinwon777: I added a cancelled error code since we see that one a lot and think it'd be helpful to distinguish cancelled from other user failures.

justinwon777:
> * How internal classification translates to the API layer (GRPC codes and rich error object). GRPC defines its own codes which cannot be changed or extended, typically observable by common infrastructure (i.e. muttley).

Added the gRPC code mapping to the error code definitions

docs/errors/errors.md
```diff
@@ -0,0 +1,239 @@
...
+```go
+var (
+	// FailureSourceUnknown is used as a fallback when no other source applies.
+	FailureSourceUnknown FailureSource = source("unknown")
+	// FailureSourceGit represents failures in git operations.
+	FailureSourceGit FailureSource = source("git")
+	// FailureSourceBazel represents failures in bazel operations.
+	FailureSourceBazel FailureSource = source("bazel")
+	// FailureSourceITG represents failures in ITG cache, changeanalyzer, and graph.
+	FailureSourceITG FailureSource = source("itg")
```
sywhang: can we clarify what ITG is. That's opaque to end users. Also what is "changeanalyzer" and "graph".

docs/errors/errors.md
```diff
@@ -0,0 +1,239 @@
...
+	// FailureSourceStorage represents failures in storage.
+	FailureSourceStorage FailureSource = source("storage")
```
sywhang: similarly, what does that mean

docs/errors/errors.md
```diff
@@ -0,0 +1,239 @@
...
+	FailureSourceOrchestrator FailureSource = source("orchestrator")
+	// FailureSourceRepoManager represents failures in repo manager.
+	FailureSourceRepoManager FailureSource = source("repomanager")
+)
+```
+
+### type TangoError
+
+```go
+type TangoError struct {
```
sywhang, line 84: Do you want to implement fmt.Formatter for this type? See https://pkg.go.dev/github.com/pkg/errors#hdr-Formatted_printing_of_errors. 

If you impl this, the errors will get a Verbose field populated using the formatter when you log it thru Zap.

docs/errors/errors.md
sbalabanov: Is there an example when UNKNOWN is classified to INTERNAL?
What is INTERNAL exactly?

docs/errors/errors.md
```diff
@@ -0,0 +1,239 @@
...
+```go
+type ErrorCode int
+```
+
+`ErrorCode` mirrors the proto `ErrorCode` enum and classifies a `TangoError` as a user error or an infra error (retryable or not).
+
+```go
+const (
+	// ErrorUnknown represents unexpected errors.
+	ErrorUnknown ErrorCode = iota
```
sbalabanov: if numbers will become a part of the contract than explicit values are better than iota

docs/errors/errors.md
```diff
@@ -0,0 +1,239 @@
...
+	return nil
+}
+```
+### Recording failure metrics
+
+In metrics package
+```go
+func recordFailure(e Emitter, op string, err error) {
+	code := tangoerrors.GetErrorCode(err)
+	source := tangoerrors.GetFailureSource(err)
```
sbalabanov: nit: most likely it will require to traverse error chain twice (once per function) so better to do in one function?

docs/errors/errors.md
```diff
@@ -0,0 +1,239 @@
...
+```go
+func GetFailureSource(err error) FailureSource
+```
+
+GetFailureSource extracts the `FailureSource` from err. If err wraps a `TangoError`, its source is returned. Otherwise `FailureSourceUnknown` is returned.
+
+## Usage
+
+### Wrapping errors at the failure site
+Callers should wrap their errors with TangoError using the provided constructors. It is up to the caller to handle classifying an error. In the following example, the bazel package creates a helper func to create a TangoError with `ErrorInfra` classification. This helper is used in `ensureBazelisk` which runs os operations that can fail with infra failures.
```
sbalabanov: potential counter case
let's say two very independent functions use storage function Get()
let's say storage implementation is mysql, and it returns mysql-specific "connection failure" error code as mysql error object, which is infra_retryable error
will both call sites run an independent classification of such an event? How do they know they have to classify mysql exit codes and cast to mysql error type?

docs/errors/errors.md
justinwon777: INTERNAL is INFRA now and it's basically all classified errors within Tango APIs that we know are not USER. UNKNOWN is only used at the controller layer. If an error reaches controller unclassified, we emit the failure with UNKNOWN and map the error to UNKNOWN for the proto response.

docs/errors/errors.md
justinwon777: cache, changeanalyzer, and graph are the subpackages in ITG. Removed it to prevent confusion.

docs/errors/errors.md
justinwon777: I would expect `Get()` to classify the error, so the two independent functions receive the classified error. I don't think it should be the caller's responsibility to know how to handle storage errors. Same with when SQ calls Tango it'd be better if Tango returned the classified error than SQ classifying it right? Which is why we're making this errors package.

docs/errors/errors.md
justinwon777: Considering we're going to log errors with the zap fields, the error, failure source, and error code will be in the log, so I don't think having another verbose field would add anything not already in the log.

docs/errors/errors.md
justinwon777: Following our discussion I removed unknown and default is now INFRA

docs/errors/errors.md
justinwon777: We don't use the values of these constants for anything so it should be fine to use iota

docs/errors/errors.md
justinwon777: GetFailureSource() has been removed

docs/errors/errors.md
justinwon777: Following our discussion I updated it so component packages will return sentinels and top level layers that call the component will check for the sentinel and classify it.

docs/errors/errors.md
```diff
@@ -0,0 +1,194 @@
...
+	...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("%w: %w", ErrDownloadBazelisk, err)
+	}
+	...
+}
+
+// orchestrator
+func classifyBazelClientError(err error) error {
```
justinwon777, line 164: A little concerned if different layers call the same component then each layer needs to implement classifying the same sentinels. I haven't seen a case yet where Tango would classify the same sentinel differently, so the classifying logic could get redundant. Although I think there aren't that many user/infra-retryable errors in Tango, so it's not like we're checking sentinels on every single error.

docs/errors/errors.md
```diff
@@ -0,0 +1,194 @@
...
+    ErrorCode code = 1;
+    String message = 2;
+}
+```
+
+| Error Code | Definition |
+|---|---|
+| `ERROR_INFRA` | Infra failures within Tango APIs not caused by the user |
+| `ERROR_CANCELLED` | Client canceled the request |
+| `ERROR_USER` | Request validation errors (missing arg, invalid arg) |
```
sbalabanov: not just validation

docs/errors/errors.md
```diff
@@ -0,0 +1,194 @@
...
+
+GetErrorCode extracts the `ErrorCode` from err. If err is `context.Canceled`, `ErrorCancelled` is returned. If err wraps a `TangoError`, its code is returned. Otherwise `ErrorInfra` is returned.
+
+## Usage
+
+The constructors are meant to be used only in top level layers (controller, orchestrator, graphrunner) that call components (git, bazel, targethasher, storage, etc). The components used by these layers are responsible for providing sentinel errors if they can be classified as user or infra-retryable. Plain errors will be classified as infra by default in the controller.
+
+```go
+// bazel
+var ErrDownloadBazelisk = errors.New("download bazelisk")
```
sbalabanov: I do not think the sentinel is a good example as it does not show the reason of the error.
HTTP download can fail because connection to a server fails (likely recoverable) or if the required version does not exist on a server (404) - likely not recoverable.
Here both of them are combined.

docs/errors/errors.md
```diff
@@ -0,0 +1,194 @@
...
+// controller
+func (c *controller) GetChangedTargets(...) error {
+	...
+	if err != nil {
+		return mapper.ToProtoError(err)
+	}
+	return nil
+}
+```
+### Recording failure metrics
```
sbalabanov: let's omit recording for now

sbalabanov (changes requested):

docs/errors/errors.md
justinwon777: Sure

docs/errors/errors.md
justinwon777: Yeah I only put validation here cause currently in OSS we only have validation errors, but in internal service we will have other possible user failures like build failures. Updated so it doesn't seem like it's only validation failures.

docs/errors/errors.md
justinwon777: The [doc](https://pkg.go.dev/net/http#Client.Do) for Do says it can return a network error and non 2xx error codes are not returned in err, but in the response body with nil err. I added a net.Error check to the example.

sbalabanov (approved):

PR #162, opened 2026-07-08
Add MaxGoroutines field to ServiceConfig to put a hard limit to the number of goroutines spawned

yushan8:
Add a configurable goroutine limit that rejects work when the process has too many goroutines, preventing runaway resource consumption. Add the check throughout the `GetChangedTargets` before we spawning goroutines.

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -16,6 +16,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"runtime"
 	"time"
@@ -38,15 +39,11 @@ type Params struct {
 	Orchestrator  orchestrator.Orchestrator
 	Scope         tally.Scope
 	ChunkConfig   config.ChunkConfig `optional:"true"`
-	MaxGoroutines int                `optional:"true"`
+	MaxGoroutines int
 }
 
 // _totalDurationBuckets covers 0–15 minutes in 10-second linear intervals.
 var _totalDurationBuckets = tally.MustMakeLinearDurationBuckets(10*time.Second, 10*time.Second, 90)
-
-const (
-	_defaultMaxGoroutines = 1000
-)
 
 type controller struct {
 	logger                 *zap.Logger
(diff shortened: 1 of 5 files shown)


controller/controller.go
```diff
@@ -76,6 +84,10 @@ func NewController(appCtx context.Context, p Params) pb.TangoYARPCServer {
 	if metadataMapChunkSize <= 0 {
 		metadataMapChunkSize = common.DefaultMetadataMapChunkSize
 	}
+	maxGoroutines := p.MaxGoroutines
+	if maxGoroutines <= 0 {
```
sbalabanov, line 85: fail fast instead of silent fallback?

controller/getchangedtargets_test.go
```diff
@@ -151,6 +151,26 @@ func TestCompareTargetGraphs(t *testing.T) {
...
+
+	request := &pb.GetChangedTargetsRequest{
+		FirstRevision:  &pb.BuildDescription{Remote: "repo:go-code", BaseSha: "sha1"},
+		SecondRevision: &pb.BuildDescription{Remote: "repo:go-code", BaseSha: "sha2"},
+		BypassCache:    true,
+	}
+
+	err := c.GetChangedTargets(request, stream)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "goroutine limit exceeded")
```
sbalabanov: do not assert on error messages

sbalabanov (changes requested):
Not sure what problem we are solving?

sywhang (changes requested):
i think you misunderstood the task description here - what we want isn't capping the total # of goroutine in the process. you want to cap the workers in repomanager's pool: https://github.com/uber/tango/blob/main/core/repomanager/repo_manager.go.

yushan8:
Ah sorry I thought the ensure no unbounded goroutine growth was related to all goroutines. Will fix.

PR #164, opened 2026-07-09
Fix path traversal issue with storage key

sywhang:
Right now we allow arbitrary paths to be passed as the storage key in all over the place with storage layer. This lets users escape the root path with "../" prefixes which can be a security vulnerability.

Fix this by validating all input sites.

### DIFF
--- a/core/storage/disk/disk.go
+++ b/core/storage/disk/disk.go
@@ -17,10 +17,13 @@ package disk
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
 	"io"
+	"io/fs"
 	"os"
-	"path/filepath"
+	"path"
 	"strings"
 
 	"github.com/uber/tango/core/storage"
@@ -49,8 +52,16 @@ func (d *diskStorage) Get(ctx context.Context, req storage.DownloadRequest) (sto
 	if ctx.Err() != nil {
 		return storage.DownloadResponse{}, ctx.Err()
 	}
-	path := filepath.Join(d.rootDir, req.Key)
-	file, err := os.Open(path)
+	if err := storage.ValidateKey(req.Key); err != nil {
+		return storage.DownloadResponse{}, err
+	}
+	root, err := os.OpenRoot(d.rootDir)
+	if err != nil {
+		return storage.DownloadResponse{}, err
+	}
+	defer root.Close()
+
+	file, err := root.Open(req.Key)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return storage.DownloadResponse{}, &storage.NotFoundError{Path: req.Key}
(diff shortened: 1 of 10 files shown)


core/storage/storage_test.go
```diff
@@ -150,3 +174,78 @@ func TestIsNotFound(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		prefix  string
+		wantErr bool
+	}{
+		{name: "Empty"},
+		{name: "PartialSegment", prefix: "graphs/repo"},
+		{name: "TrailingSlash", prefix: "graphs/repo/"},
+		{name: "DotPrefix", prefix: "."},
+		{name: "DotDotPrefix", prefix: ".."},
```
xytan0056: this should be have `wantErr == true`?

sywhang: dotdot's okay, but traversing is not okay. see L225

core/storage/disk/disk.go
```diff
@@ -89,15 +108,45 @@ func (d *diskStorage) Put(ctx context.Context, req storage.UploadRequest) error
 	if err := tmp.Close(); err != nil {
 		return err
 	}
-	return os.Rename(tmpPath, path)
+	return root.Rename(tmpPath, req.Key)
+}
+
+func createTemp(root *os.Root, dir string) (*os.File, string, error) {
+	for range 10 {
```
sbalabanov: why need to retry if you can instead just pick arbitrary long random sequence?

core/storage/disk/disk.go
```diff
@@ -89,15 +108,45 @@ func (d *diskStorage) Put(ctx context.Context, req storage.UploadRequest) error
 	if err := tmp.Close(); err != nil {
 		return err
 	}
-	return os.Rename(tmpPath, path)
+	return root.Rename(tmpPath, req.Key)
+}
+
+func createTemp(root *os.Root, dir string) (*os.File, string, error) {
```
sbalabanov, line 121: deserves a comment why `os.CreateTemp` isn't used

core/storage/storage.go
```diff
@@ -34,6 +39,57 @@ func IsNotFound(err error) bool {
...
+	return fmt.Sprintf("invalid storage key %q", e.Key)
+}
+
+// IsInvalidKey reports whether err identifies an invalid storage key.
+func IsInvalidKey(err error) bool {
+	var target *InvalidKeyError
+	return errors.As(err, &target)
+}
+
+// ValidateKey validates a storage key.
```
sbalabanov: may be explain how

sbalabanov (changes requested):
sending back as per offline discussion

core/storage/storage.go
sywhang: this change is undone.

sywhang:
discussed offline - upon another thought: the disk storage layer itself needs to treat the keys as a path-unaware thing per contract defined in our storage layer. Let's just close this PR and fix disk storage implementation to treat the keys as true opaque strings.

PR #168, opened 2026-07-09
add internal/targetdiff

sywhang:
This is part of refactoring the core package structures to make it more sensible.

Currently, our controller layer is not really a controller - that's a full on handler that bundles up the proto RPC request layer + what looks like should be 3 different controller methods:
- change graph computation thru bazel
- graph caching
- comparison

all bundled up in to a single method: GetChangedTargets.

Not only that violates the service architecture, but also this whole thing makes it difficult to test and leaks RPC abstraction into various parts of the logic.

Let's begin by creating a separate package for graph comparison.

internal/targetdiff is an internal pkg that lets us compare targets across two changed targets graph.

### DIFF
--- a/internal/targetdiff/compare.go
+++ b/internal/targetdiff/compare.go
@@ -23,14 +23,22 @@ const _sourceFileRuleType = "source file"
 
 // Target describes a build target using semantic names.
 type Target struct {
-	Name         string
-	Hash         string
-	RuleType     string
+	// Name is the target's canonical build label.
+	Name string
+	// Hash identifies the target's state, including relevant dependencies.
+	Hash string
+	// RuleType identifies the kind of build rule or source file.
+	RuleType string
+	// Dependencies contains the names of the target's direct dependencies.
 	Dependencies []string
-	Tags         []string
-	Attributes   map[string]string
-	Root         bool
-	External     bool
+	// Tags contains the target's build tags.
+	Tags []string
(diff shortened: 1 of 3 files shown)


xytan0056 (approved):

internal/targetdiff/compare.go
```diff
@@ -0,0 +1,291 @@
...
+
+import "context"
+
+const _cancelCheckInteral = 4096
+
+const _sourceFileRuleType = "source file"
+
+// Target describes a build target using semantic names.
+type Target struct {
+	Name         string
```
sbalabanov: plz comment all exported fields

internal/targetdiff/compare.go
```diff
@@ -0,0 +1,291 @@
...
+	ChangeTypeInvalid ChangeType = iota
+	ChangeTypeNew
+	ChangeTypeDeleted
+	ChangeTypeChanged
+)
+
+// ChangedTarget describes one target that differs between revisions.
+type ChangedTarget struct {
+	ChangeType ChangeType
+	Before     *Target
```
sbalabanov: should it be a pointer?

sbalabanov (approved):

internal/targetdiff/compare.go
sywhang: yes, it has a semantic. the docs change should clarify this.

PR #169, opened 2026-07-09
feat(errors): create core/errors package

justinwon777:
Create the errors package following the [RFC](https://github.com/uber/tango/blob/main/docs/errors/errors.md).

## Test Plan
unit tests


core/errors/errors.go
```diff
@@ -0,0 +1,172 @@
...
+		return "infra"
+	case ErrorInfraRetryable:
+		return "infra_retryable"
+	default:
+		return "unknown"
+	}
+}
+
+// FailureSource represents the component where an error occurred.
+type FailureSource interface {
```
sywhang: let's add some more docs around this being an sealed interface type. Specifically, we should add docs clarifying that this is sealed because we want to lock down the metrics cardinality that can be added to this, and don't want arbitrary failure source types to be thrown around everywhere.

sywhang (approved):

PR #171, opened 2026-07-09
refactor(logging): replace log.Printf with injected zap logger and remove dead code

xytan0056:
## What
replace `log.Printf` with universal zap logger and thread the logger to call sites required
removed unused functions in ITG

## Test plan
CI tests

### DIFF
--- a/core/git/git.go
+++ b/core/git/git.go
@@ -21,7 +21,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
 	"strconv"
@@ -29,6 +28,7 @@ import (
 	"time"
 
 	"github.com/uber/tango/core/execcmd"
+	"go.uber.org/zap"
 )
 
 const (
@@ -63,13 +63,19 @@ type Interface interface {
 type impl struct {
 	directory string
 	runner    commandRunner
+	logger    *zap.SugaredLogger
 }
 
-// New creates new Git interface implementation
-func New(directory string) Interface {
+// New creates new Git interface implementation. A nil logger is tolerated and
+// discards log output.
+func New(directory string, logger *zap.SugaredLogger) Interface {
+	if logger == nil {
+		logger = zap.NewNop().Sugar()
+	}
 	return &impl{
 		directory: directory,
 		runner:    &osExecRunner{},
+		logger:    logger,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


core/git/git.go
```diff
@@ -63,13 +63,19 @@ type Interface interface {
...
 	runner    commandRunner
+	logger    *zap.SugaredLogger
 }
 
-// New creates new Git interface implementation
-func New(directory string) Interface {
+// New creates new Git interface implementation. A nil logger is tolerated and
+// discards log output.
+func New(directory string, logger *zap.SugaredLogger) Interface {
+	if logger == nil {
```
xytan0056, line 72: another option is to extract the logger from ctx, but I think it's more at the request scope, rather than for util module

core/itg/graph/graph.go
```diff
@@ -238,14 +238,18 @@ func (g *OptimizedGraph) AddTarget(target *targethasher.Target) {
 }
 
 // GetReverseDepsAsTargets returns the reverse dependencies of the given targets.
-func (g *OptimizedGraph) GetReverseDepsAsTargets(targetNames []string) []targethasher.Target {
+// A nil logger is tolerated and discards log output.
+func (g *OptimizedGraph) GetReverseDepsAsTargets(logger *zap.SugaredLogger, targetNames []string) []targethasher.Target {
```
xytan0056: adding a logger in OptimizedGraph is weird. cuz it's a cache object to be marshaled, but I indeed think this warn message should be printed for stuff like exported non-existing files, as warning but not failure. So threading it in.

core/itg/graph/graph.go
```diff
@@ -238,14 +238,17 @@ func (g *OptimizedGraph) AddTarget(target *targethasher.Target) {
...
+		logger = zap.NewNop().Sugar()
+	}
 	reverseDeps := make([]targethasher.Target, 0, 2*len(targetNames))
 	visited := NewIntSet()
 	reverseDepIDs := make([]int, 0, len(targetNames))
 	for _, targetName := range targetNames {
 		targetID, ok := g.TargetNameToID[targetName]
 		if !ok {
-			log.Printf("starting target %s is not in the graph", targetName)
+			logger.Warnw("starting target is not in the graph", "target", targetName)
```
sywhang: if this call site is the only place we're logging something...... but hang on, do we want to log here or explicitly error out? when does this even occur?

core/itg/graph/graph.go
xytan0056: dead code not needed

core/itg/graph/graph.go
xytan0056: good catch . actually this is dead code.
removed.

core/git/git.go
sywhang: ideally we should just move this entire pkg to internal. this looks fine for now.

sywhang (approved):

PR #172, opened 2026-07-10
Ensure all configs are used or checked in the constructor

yushan8:
All config validation is now done in constructors. Methods no longer check for config values. Return a validation error if the config was not set.


config/config.go
```diff
@@ -70,12 +70,24 @@ func Parse(configFilePath string) (*Config, error) {
...
+	}
 	config.repositoryByRemote = make(map[string]*RepositoryConfig, len(config.Repository))
 	for i := range config.Repository {
 		remote := config.Repository[i].Remote
 		if remote == "" {
 			return nil, fmt.Errorf("repository[%d].remote must not be empty", i)
 		}
+		if config.Repository[i].QueryTimeoutSeconds <= 0 {
+			return nil, fmt.Errorf("repository[%d].query_timeout_seconds must be > 0", i)
+		}
```
liujsj, line 90: Why are we removing the default values and erroring out. Should we not have default values if user do not provide for timeout or chunks? 

I am not sure what the config document will look like but initial gut reaction is that we are asking users of the library to be quite knowledgable in the code base vs working out of the box

liujsj: Also if we do decide to change that these must be provided, then please update service_config comments as it mentions

// streaming chunk sizes; zero values fall back to package defaults

core/bazel/query_test.go
```diff
@@ -57,6 +57,7 @@ func TestExecuteQuery_Success(t *testing.T) {
 		mockCmd.EXPECT().Wait().Return(nil),
 	)
 	client, err := NewBazelClient(context.Background(), Params{
+		QueryTimeout:  5 * time.Minute,
```
sbalabanov, line 60: it is better to rely on test timeout rather than hardcode

sbalabanov (approved):

yushan8:
Will update once config rfc is up. Moving change for checking the default values in methods and ensuring they're used in https://github.com/uber/tango/pull/175

PR #173, opened 2026-07-10
migrate controller to core/errors TangoError

justinwon777:
This removes all usages of core/common errors in controller and classifies user failures with TangoError.
`GetChangedTargets` and `GetTargetGraph` return error with `mapper.ToProtoError(retErr)` so all errors are classified.

## Test Plan

unit tests

### DIFF
--- a/controller/gettargetgraph.go
+++ b/controller/gettargetgraph.go
@@ -16,13 +16,12 @@ package controller
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"time"
 
 	"github.com/uber/tango/core/cachekey"
-	"github.com/uber/tango/core/common"
+	tangoerrors "github.com/uber/tango/core/errors"
 	"github.com/uber/tango/entity"
 	"github.com/uber/tango/internal/mapper"
 	"github.com/uber/tango/internal/url"
@@ -44,6 +43,7 @@ func (c *controller) GetTargetGraph(request *pb.GetTargetGraphRequest, stream pb
 			logger.Error("GetTargetGraph failed", zap.Error(retErr))
 			scope.Counter("failure").Inc(1)
 			emitFailureMetric(scope, retErr)
+			retErr = mapper.ToProtoError(retErr)
 		} else {
 			scope.Counter("success").Inc(1)
 		}
@@ -53,12 +53,12 @@ func (c *controller) GetTargetGraph(request *pb.GetTargetGraphRequest, stream pb
 	defer cancelLink()
 	entityReq, err := mapper.ProtoToGetTargetGraphRequest(request)
 	if err != nil {
-		return fmt.Errorf("convert get target graph request: %w", err)
+		return tangoerrors.NewUser(fmt.Errorf("convert get target graph request: %w", err))
 	}
 	scope = scope.Tagged(map[string]string{"repo": url.ToShortRemote(entityReq.Build.Remote)})
 	graphReader, err := c.getGraph(ctx, entityReq)
 	if err != nil {
-		return err
+		return fmt.Errorf("get graph: %w", err)
 	}
 	if graphReader == nil {
 		// Nothing to stream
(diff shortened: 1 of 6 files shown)


controller/gettargetgraph.go
```diff
@@ -126,10 +127,10 @@ func (c *controller) getGraph(ctx context.Context, req entity.GetTargetGraphRequ
 			graphReader, err := storage.NewGraphReader(ctx, c.storage, treehashPath)
 			if err != nil {
 				if ctx.Err() != nil {
-					return nil, common.WithReason(common.FailureReasonCancelled, common.ErrorTypeUser, ctx.Err())
+					return nil, ctx.Err()
 				}
 				if !storage.IsNotFound(err) {
-					return nil, common.WithReason(failureReasonGraphFetch, common.ErrorTypeInfra, err)
+					return nil, err
```
yushan8: nit: Error should be wrapped with `fmt.Errorf("graph reader: %w", err)`

controller/gettargetgraph.go
```diff
@@ -151,13 +152,9 @@ func (c *controller) getGraph(ctx context.Context, req entity.GetTargetGraphRequ
...
 		if ctx.Err() != nil {
-			return nil, common.WithReason(common.FailureReasonCancelled, common.ErrorTypeUser, ctx.Err())
+			return nil, ctx.Err()
 		}
-		var ce common.ClassifiedError
-		if errors.As(err, &ce) {
-			return nil, err
-		}
-		return nil, common.WithReason(failureReasonGraphFetch, common.ErrorTypeInfra, err)
+		return nil, err
```
yushan8: Same here error should be wrapped.

controller/gettargetgraph.go
```diff
@@ -80,12 +81,12 @@ func (c *controller) GetTargetGraph(request *pb.GetTargetGraphRequest, stream pb
 			return nil
 		}
 		if err != nil {
-			return common.WithReason(failureReasonGraphFetch, common.ErrorTypeInfra, err)
+			return err
```
yushan8: Nit: wrap error on `graphReader read`

controller/gettargetgraph.go
justinwon777: yeah good idea since these are losing the failure reasons.

controller/gettargetgraph_test.go
```diff
@@ -376,10 +369,7 @@ func TestGetTargetGraph_OrchestratorCancelled(t *testing.T) {
 		BuildDescription: &pb.BuildDescription{Remote: "repo:go-code", BaseSha: "sha"},
 	}, stream)
 	require.Error(t, err)
-	var ce common.ClassifiedError
-	require.True(t, errors.As(err, &ce))
-	assert.Equal(t, common.FailureReasonCancelled, ce.Reason())
-	assert.Equal(t, common.ErrorTypeUser, ce.Type())
+	assert.Contains(t, err.Error(), "context canceled")
```
yushan8: Nit: don't assert on error string

controller/gettargetgraph.go
```diff
@@ -151,13 +152,9 @@ func (c *controller) getGraph(ctx context.Context, req entity.GetTargetGraphRequ
 	graphReader, err := c.orchestrator.GetTargetGraph(ctx, req)
 	if err != nil {
 		if ctx.Err() != nil {
-			return nil, common.WithReason(common.FailureReasonCancelled, common.ErrorTypeUser, ctx.Err())
+			return nil, ctx.Err()
```
yushan8: Should we wrap the context cancelled errors too so we know where it came from?

controller/gettargetgraph_test.go
justinwon777: Removed error string assertions. That's because asserting on error strings we don't control is fragile?

yushan8: yea I think with it's better to use `errors.As` if it's a certain type or checking on exit code if it's provided.

yushan8 (approved):

PR #174, opened 2026-07-10
refactor(controller): use internal/targetdiff for GetChangedTargets

xytan0056:
Now that we have `internal/compare.go`, move the diff/classify/BFS logic in `getchangedtargets.go` to `internal/targetdiff.Compare`. The controller decodes each graph stream into a semantic graph, compares, and re-maps. Behavior not changed.

- make getchangedtargets Controller call internal/targetdiff
- break getchangedtargets to component functions

## Test
make test
Manually tested locally
```
make run-client-changed-targets REMOTE=https://github.com/uber/tango.git BASE_SHA=74d1cd55155e5f4f43aa92b4e0146a0c528a0d96  NEW_BASE_SHA=74d1cd55155e5f4f43aa92b4e0146a0c528a0d96~
```
```
//controller:controller_test: {"distance":3}

bazel query "somepath(//controller:controller_test , //orchestrator:native_orchestrator.go )"
//controller:controller_test
//orchestrator/orchestratormock:orchestratormock
//orchestrator:orchestrator
//orchestrator:native_orchestrator.go
```


controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+		// Blob is corrupt (likely an incomplete write). Log and fall through to recompute.
+		logger.Warn("GetChangedTargets: Cached result is incomplete, recomputing", zap.Error(readErr))
+		return false, nil
+	}
+
+	cacheReadDuration := time.Since(cacheStart)
+	logger.Info("GetChangedTargets: Cache hit, streaming from storage",
+		zap.Duration("cache_read_duration", cacheReadDuration),
+	)
+	scope.Counter("changed_targets_cache_hit").Inc(1)
```
xytan0056, line 194: keep the old metric recording here until the metrics package is ready

controller/getchangedtargets.go
xytan0056, line 48: now the call flow is 
GetChangedTargets 
-> serveChangedTargetsFromCache 
-> fetchTargetGraphs for both revs 
-> compareTargetGraphs 
-> transpose 
-> cacheComparedTargets

controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
 		maxDist = request.GetOutputConfig().GetMaxDistance()
 	}
```
liujsj, line 80: Handler usually does input validation. Since the code is both the controller and handler, here if user provides anything less than 0 then we shouldn't we error out right away?

controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
-	// (e.g. a missing-deadline "missing TTL" reject) becomes a visible request failure
-	// rather than silent degradation.
+	// Fast path: stream a previously computed result straight from cache.
 	if !request.GetBypassCache() {
-		cacheStart := time.Now()
-		treehash1, treehash2, err := readTreehashParallel(ctx, c.storage, request.GetFirstRevision(), request.GetSecondRevision())
-		if err != nil {
-			logger.Error("GetChangedTargets: Failed to read revision treehash", zap.Error(err))
-			return common.WithReason(failureReasonTreehashRead, common.ErrorTypeInfra, err)
+		served, err := c.serveChangedTargetsFromCache(ctx, scope, logger, request, stream, maxDist, start)
```
liujsj, line 84: Just to confirm, with the refactoring to scope and logger, we won't need to provide this as input everywhere (Initially I was wondering why aren't these set as fields on the controller where we can do c.scope(), c.logger()

controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+		if err != nil || served {
+			return err
 		}
```
liujsj: This is somewhat confusing because I would expect some differences between whether there was an error vs we completed the request from using cached information. 

If we do serve the request from the cache, shouldn't we emit a metric so the code flow should be different?

controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+			return common.WithReason(common.FailureReasonCancelled, common.ErrorTypeUser, ctx.Err())
 		}
+		logger.Error("GetChangedTargets: Failed to compare target graphs", zap.Error(err))
+		return common.WithReason(failureReasonCompare, common.ErrorTypeInfra, fmt.Errorf("failed to compare target graphs: %w", err))
 	}
+	compareDuration := time.Since(compareStart)
+	logger.Info("GetChangedTargets: Target graphs compared",
+		zap.Duration("compare_duration", compareDuration),
+	)
+	scope.Timer("compare_duration").Record(compareDuration)
```
liujsj: Can we be more consistent in how we add logs and emit metrics. For example in the helper method, fetchTargetGraphs, it was responsible for logging and emitting metrics. But now for compareTargetGraphs, it is on the top level method to handle this.

controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
 
-	jobs := make([]*job, 2)
+	// Cache the computed result concurrently so it doesn't block the stream send.
+	c.cacheComparedTargets(logger, request, changedTargetsResponses)
```
liujsj, line 111: IMO it should be the responsibility of the top level method to decide whether to cache concurrently (In a goroutine or not). 

The underlying helper method should just do the caching comparing logic, not make decision on whether it should be in a goroutine or not

controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+
+	sendStart := time.Now()
+	if err := sendTrimmedChangedTargets(stream, changedTargetsResponses, maxDist, request.GetOutputConfig()); err != nil {
+		logger.Error("GetChangedTargets: Failed to send response", zap.Error(err))
+		return common.WithReason(failureReasonSend, common.ErrorTypeInfra, fmt.Errorf("failed to send response: %w", err))
+	}
+	sendDuration := time.Since(sendStart)
+	scope.Timer("send_duration").Record(sendDuration)
```
liujsj, line 118: When we handle the cache response, it was the responsibility for the helper method to deal with all the metrics and logs but if it isn't cached then it is the top level method. Can we be consistent here.

controller/getchangedtargets.go
```diff
@@ -73,76 +75,147 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+		zap.Duration("compare_duration", compareDuration),
+	)
+	scope.Timer("compare_duration").Record(compareDuration)
 
-	jobs := make([]*job, 2)
+	// Cache the computed result concurrently so it doesn't block the stream send.
+	c.cacheComparedTargets(logger, request, changedTargetsResponses)
+
+	sendStart := time.Now()
+	if err := sendTrimmedChangedTargets(stream, changedTargetsResponses, maxDist, request.GetOutputConfig()); err != nil {
```
liujsj, line 114: Is it possible to merge the two logics of sending the result via stream regardless of it is cached or not. My initial thought is that regardless on whether we grabbed the result from computing or caching, the underlying way we send the response back should be some shared method. 

Right now it looks like sendTrimmedChangedTargets and servceChangedTargetsFromCache has their own logic of sending the response back

controller/getchangedtargets.go
xytan0056: <0 is deliberate signal for [no filtering](url). 
but other than that, imo config erros should be gated in #172 once the config contract is settled, so at this point GetMaxDistance() should be sane

controller/getchangedtargets.go
xytan0056: yes, these shouldn't have been passed as arguments -- will be further refactored once metrics package is ready to use.  We will **extract scope from ctx**. 
For now, because the ctx is not there yet, we need to [pass them in like this](https://github.com/cadence-workflow/cadence/blob/54b9b08ebedf3db4c33c3b29ffaa29381ff9b298/service/worker/scheduler/backfill.go#L34), we cannot share the scope from controller struct like `c.`, because each of these methods technically are child scope, from parent, which doesn't carry some enriched infomation.

controller/getchangedtargets.go
xytan0056: oh this is not comparing, this is to fire and forget function to cache the ctc result. so that the next request for the same treehash compare could return fast from cache
So the "goroutine or not" decision is supposed to be determined in the underlying function (reason being use separte appCtx + best effort + caller not changing response, these re nothing to do with caller)

controller/getchangedtargets.go
xytan0056: they are the same sendTrimmedChangedTargets call
```
if cache {
    // internally calls sendTrimmedChangedTargets
    serveCache
    return
} 
// compute...
sendTrimmedChangedTargets
return 
```

controller/getchangedtargets.go
xytan0056: metric is emitted inside serveChangedTargetsFromCache.

readability is a good point. collapsed if clause

controller/getchangedtargets.go
xytan0056: same as above. moved.

controller/getchangedtargets.go
xytan0056: good point, unified all metrics emission to inside.

controller/getchangedtargets.go
```diff
@@ -303,40 +358,21 @@ func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, str
...
+// GetChangedTargetsResponse stream. Each stream is decoded into a semantic
+// targetdiff.Graph (int32 IDs resolved to names via that stream's metadata),
+// the two graphs are compared by internal/targetdiff, and the resulting changes
+// are re-mapped into a canonical per-call ID namespace so the response metadata
+// only carries the names actually referenced. See internal/targetdiff for the
+// classification and distance rules.
+func (c *controller) compareTargetGraphs(ctx context.Context, scope tally.Scope, logger *zap.Logger, firstGraph, secondGraph []*pb.GetTargetGraphResponse, maxDist int32) ([]*pb.GetChangedTargetsResponse, error) {
 	start := time.Now()
-	scope := c.scope.SubScope("compare_target_graphs")
+	compareScope := c.scope.SubScope("compare_target_graphs")
```
sywhang: a little strange? this takes in `scope` parameter but you're using the scope defined in the controller.

controller/getchangedtargets_test.go
```diff
@@ -1517,3 +1379,204 @@ func TestSendTrimmedChangedTargets_RetainsDeletedAtMaxDistanceOne(t *testing.T)
...
+			func(_ context.Context, _ orchestrator.GetTargetGraphParam) (storage.GraphReader, error) {
+				return newMockGraphReader(ctrl), nil
+			}).Times(2)
+
+		c := newTestController(zaptest.NewLogger(t))
+		c.orchestrator = orch
+
+		_, _, err := c.fetchTargetGraphs(context.Background(), c.scope, c.logger, bypassRequest())
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "no chunks returned")
```
sywhang: let's not assert on error msgs.

controller/getchangedtargets_test.go
```diff
@@ -1517,3 +1379,204 @@ func TestSendTrimmedChangedTargets_RetainsDeletedAtMaxDistanceOne(t *testing.T)
...
+		orch := orchestratormock.NewMockOrchestrator(ctrl)
+		orch.EXPECT().GetTargetGraph(gomock.Any(), gomock.Any()).DoAndReturn(
+			func(_ context.Context, _ orchestrator.GetTargetGraphParam) (storage.GraphReader, error) {
+				panic("boom in orchestrator")
+			}).Times(2)
+
+		c := newTestController(zaptest.NewLogger(t))
+		c.orchestrator = orch
+
+		_, _, err := c.fetchTargetGraphs(context.Background(), c.scope, c.logger, bypassRequest())
```
sywhang: generic comment: we'er using bg ctx throughout the test. let's use t.Context().

controller/getchangedtargets.go
```diff
@@ -606,218 +527,68 @@ func getTargetsAndMetadata(ctx context.Context, graph []*pb.GetTargetGraphRespon
 	return targets, merged, nil
 }
 
-// buildNameIndex creates name->target maps using the provided metadata information.
-func buildNameIndex(ctx context.Context, targetsByID map[int32]*pb.OptimizedTarget, meta *pb.Metadata) (map[string]*pb.OptimizedTarget, error) {
-	byName := make(map[string]*pb.OptimizedTarget, len(targetsByID))
+// toDiffGraph resolves a stream's int32 IDs into a semantic targetdiff.Graph
+// keyed by canonical target name. Targets whose ID has no name mapping are
+// skipped, mirroring the prior name-index behavior; dependency, tag, and
```
sywhang: let's add a test that asserts this behavior

controller/getchangedtargets.go
xytan0056: ah good catch, the refactor was purely cosmetic

controller/getchangedtargets.go
xytan0056: good point

controller/getchangedtargets.go
liujsj: I still see some inconsistency and also code duplication

```suggestion
	totalDuration := time.Since(start)
	logger.Info("GetChangedTargets: Successfully processed request",
		zap.Duration("send_duration", sendDuration),
		zap.Duration("total_duration", totalDuration),
	)
	scope.Timer("total_duration").Record(totalDuration)
	scope.Histogram("total_duration.histogram", c.totalDurationBuckets).RecordDuration(totalDuration)
```

This is in the top level handler/controller method but also in `serveChangedTargetsFromCache`. Can we remove the duplicate code somehow

sywhang (approved):

PR #175, opened 2026-07-10
Make sure config values are actually being used not defaults

yushan8:
A couple places in the mapper and chunking is not using config values and using the defaults. Make sure config values are actually being used not defaults.

### DIFF
--- a/core/common/utils.go
+++ b/core/common/utils.go
@@ -244,6 +244,10 @@ func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Res
 }
 
 func chunkTargets(targets []*tangopb.OptimizedTarget, chunkSize int) []*tangopb.GetTargetGraphResponse {
+	if chunkSize <= 0 {
+		chunkSize = DefaultTargetChunkSize
+	}
+
 	// at least one chunk
 	numChunks := max(1, (len(targets)+chunkSize-1)/chunkSize)
 
@@ -291,6 +295,10 @@ func ChunkMetadata(
 	attrStrValIDToVal map[int32]string,
 	chunkSize int,
 ) []*tangopb.Metadata {
+	if chunkSize <= 0 {
+		chunkSize = DefaultMetadataMapChunkSize
+	}
+
 	targetChunks := splitMap(targetIDToName, chunkSize)
 	attrValChunks := splitMap(attrStrValIDToVal, chunkSize)
 
--- a/core/common/utils.go
+++ b/core/common/utils.go
@@ -140,6 +140,7 @@ const cancelCheckInterval = 4096
 // ResultToGetTargetGraphResponse converts a Result to a GetTargetGraphResponse.
 // targetChunkSize controls how many OptimizedTarget entries per stream message.
 // metadataMapChunkSize controls how many entries per metadata map chunk.
+// TODO: move this function to internal/mapper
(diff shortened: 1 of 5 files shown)


core/common/utils.go
```diff
@@ -137,8 +137,10 @@ func HashRequestOptions(opts *tangopb.RequestOptions) string {
 // for typical target rates.
 const cancelCheckInterval = 4096
 
-// ResultToGetTargetGraphResponse converts a Result to a GetTargetGraphResponse
-func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Result) ([]*tangopb.GetTargetGraphResponse, error) {
+// ResultToGetTargetGraphResponse converts a Result to a GetTargetGraphResponse.
+// targetChunkSize controls how many OptimizedTarget entries per stream message.
+// metadataMapChunkSize controls how many entries per metadata map chunk.
+func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Result, targetChunkSize, metadataMapChunkSize int) ([]*tangopb.GetTargetGraphResponse, error) {
```
yushan8, line 144: method will be moved out of common.

core/common/utils.go
```diff
@@ -242,10 +244,6 @@ func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Res
 }
 
 func chunkTargets(targets []*tangopb.OptimizedTarget, chunkSize int) []*tangopb.GetTargetGraphResponse {
-	if chunkSize <= 0 {
-		chunkSize = DefaultTargetChunkSize
-	}
```
liujsj, line 247: If we do this, what happens now where someone does not specify a chunksize?

I was expecting if this was done here, there would be some other change where now this is handled in the constructor.

yushan8: Yea this should be handled in the constructor. I think this method should be internal/mappers so clients can't use it.

yushan8: Will revert the change for now and combine it with the constructor config validation.

core/common/utils.go
sywhang: can you leave a TODO comment

core/common/utils.go
sywhang: +1 to @liujsj 's comments. But at a high level I think we need a config validator that lives in the config pkg that gets asserted once upon fx init

liujsj (approved):

orchestrator/native_orchestrator.go
xytan0056:
```suggestion
err = ws.Checkout(ctx, remote, param.Req.BuildDescription.BaseSha)
if err != nil {
return nil, fmt.Errorf("checkout %s@%s: %w", remote, param.Req.BuildDescription.BaseSha, err)
```

xytan0056:
```suggestion
	treehashPath := common.GetGraphByTreeHash(remote, treehash, param.Req.BuildDescription.GetStrategy(), param.Req.GetRequestOptions())
```

core/common/utils_test.go
```diff
@@ -198,16 +197,45 @@ func TestResultToGetTargetGraphResponse_Chunking(t *testing.T) {
...
+	tests := []struct {
+		name                 string
+		targetChunkSize      int
+		metadataMapChunkSize int
+		wantTargetChunks     int
+	}{
+		{
+			name:                 "250 per chunk",
+			targetChunkSize:      250,
+			metadataMapChunkSize: 50_000,
```
xytan0056: nit: set it to a smaler value to test the metadata chunks size, 100-200 ish

xytan0056 (approved):

PR #177, opened 2026-07-10
feat(internal/cachekey): add cache path/key construction package

justinwon777:
Adds internal/cachekey with the cache-path builders (GetGraphByTreeHash, GetTreehashCachePath, GetComparedTargetsCachePath) that today live in core/common/utils.go. A follow-up PR will switch controller and orchestrator to use this package.

## Test Plan

unit tests

### DIFF
--- a/internal/cachekey/cachekey.go
+++ b/internal/cachekey/cachekey.go
@@ -1,0 +1,105 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package cachekey builds the storage keys used to cache target graphs, treehashes, and compared-target-graph results.
+package cachekey
+
+import (
+	"crypto/md5"
+	"fmt"
+	"path/filepath"
+	"sort"
+
+	"github.com/uber/tango/core/common"
+	"github.com/uber/tango/tangopb"
+)
+
+// GetGraphByTreeHash returns the cache path for the target graph by treehash.
+// strategy is part of the key because different computation strategies (e.g.
+// SHELL vs NATIVE) can produce different graphs from the same tree state.
+// requestOptions is folded into the key when any of its fields affect computation
+// (today: extra_exclude_files_regex). Empty/nil ⇒ legacy path unchanged.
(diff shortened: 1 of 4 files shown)


internal/cachekey/cachekey.go
```diff
@@ -0,0 +1,112 @@
...
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/uber/tango/tangopb"
+)
+
+// ToShortRemote returns the short remote name given a git ssh remote string.
+// For example, "git@github:uber/tango" will return "uber/tango".
+func ToShortRemote(remote string) string {
```
sywhang: this conflicts with some of the changes I'm trying to make in https://github.com/uber/tango/pull/164.... Can we wait till that one is merged, or at least propagate the changes here?

justinwon777: yeah I'll wait for yours to merge

justinwon777: Saw that PR closed, so you're not making changes to these then?

xytan0056:
let's update CLAUDE.md as well
https://github.com/uber/tango/blob/main/CLAUDE.md#caching-and-treehashes

internal/cachekey/cachekey.go
```diff
@@ -0,0 +1,112 @@
...
+	path := filepath.Join(ToShortRemote(remote), "compared-targets", treehash1+"_"+treehash2)
+	if hash := hashRequestOptions(requestOptions); hash != "" {
+		path += "_requests-options-" + hash
+	}
+	return path
+}
+
+// getReqURLsHash returns a fixed-length MD5 hash of the sorted request URLs.
+// Each URL's bytes are fed into the digest individually (no separator), matching
+// the Java MessageDigest.update(str.getBytes()) per-string behavior.
```
xytan0056:
```suggestion
// Each URL's bytes are fed into the digest individually (no separator)
```

internal/cachekey/cachekey.go
```diff
@@ -0,0 +1,112 @@
...
+	return strs[len(strs)-1]
+}
+
+// GetGraphByTreeHash returns the cache path for the target graph by treehash.
+// strategy is part of the key because different computation strategies (e.g.
+// SHELL vs NATIVE) can produce different graphs from the same tree state.
+// requestOptions is folded into the key when any of its fields affect computation
+// (today: extra_exclude_files_regex). Empty/nil ⇒ legacy path unchanged.
+func GetGraphByTreeHash(remote, treehash string, strategy tangopb.ComputationStrategy, requestOptions *tangopb.RequestOptions) string {
+	path := filepath.Join(ToShortRemote(remote), "graphs", treehash, strategy.String())
```
xytan0056: nit: ToShortRemote is also used in other places. (actually by accdient [duplicated here](https://github.com/uber/tango/blob/c28b82827e7ac7d5f6b6e4141fe93e04f0d30178/core/repomanager/repo_manager.go#L205))
I think this is a case where `util.ToShortRemote` is better than `cachekey.ToShortRemote`

justinwon777: Yeah that shouldn't be in here. Removed from the package.

xytan0056 (approved):

PR #178, opened 2026-07-10
feat(internal/mapper/idmapper): add ID-mapper package

justinwon777:
Adds internal/mapper/idmapper with mapper helpers that today live in core/common. A follow-up PR will switch controller and orchestrator over to it and remove the duplicated code from core/common.

## Test Plan

unit test

### DIFF
--- a/internal/mapper/idmapper/idmapper.go
+++ b/internal/mapper/idmapper/idmapper.go
@@ -1,0 +1,55 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package idmapper provides the ID-mapping primitives used to build the tangopb
+// wire format's int32-ID-keyed maps.
+package idmapper
+
+// Mapper assigns stable int32 IDs to string names on demand.
+// IDs are assigned sequentially starting from 1. Zero is reserved as the
+// proto3 "unset" sentinel so consumers using encoding/json (which honors
+// `omitempty` on int32 fields) or any client that treats GetId() == 0 as
+// missing never silently lose real entries.
+type Mapper struct {
+	nameToID map[string]int32
+	nextID   int32
+}
+
+// NewMapper creates a new Mapper.
+func NewMapper() *Mapper {
+	return &Mapper{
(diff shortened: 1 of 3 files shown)


internal/idmap/idmap.go
```diff
@@ -0,0 +1,132 @@
...
+		out[id] = name
+	}
+	return out
+}
+
+// ChunkMetadata splits the metadata maps into multiple Metadata messages.
+// target_id_mapping and attribute_string_value_mapping scale with repo size and can exceed the
+// 64MB gRPC per-message limit for large monorepos; they are split across chunks of chunkSize entries.
+// The small maps (rule_type, tag, attribute_name) always fit in one message and are sent in the first chunk.
+func ChunkMetadata(
```
liujsj: What was the reasoning to have chunking logic grouped into this file?

justinwon777: Initially cause I saw chunking is used with the idmapper so I thought they were related enough, but they have different purposes so I agree chunking should be in a different package. Removed.

internal/mapper/idmapper/idmapper.go
```diff
@@ -0,0 +1,55 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package idmapper provides the ID-mapping primitives used to build the tangopb
+// wire format's int32-ID-keyed maps.
+package idmapper
```
justinwon777, line 17: This maybe should be under `internal/mapper/idmapper`, but wasn't sure if `internal/mapper` is supposed to be only for just proto mappers.

internal/idmapper/idmapper.go
```diff
@@ -0,0 +1,55 @@
...
+// Package idmapper provides the ID-mapping primitives used to build the tangopb
+// wire format's int32-ID-keyed maps.
+package idmapper
+
+// NameIDMapper assigns stable int32 IDs to string names on demand.
+// IDs are assigned sequentially starting from 1. Zero is reserved as the
+// proto3 "unset" sentinel so consumers using encoding/json (which honors
+// `omitempty` on int32 fields) or any client that treats GetId() == 0 as
+// missing never silently lose real entries.
+type NameIDMapper struct {
```
xytan0056: nit: `idmapper.Mapper` sounds cleaner than `idmapper.NameIdMapper`?

internal/idmapper/idmapper.go
```diff
@@ -0,0 +1,55 @@
...
+	nameToID map[string]int32
+	nextID   int32
+}
+
+// NewNameIDMapper creates a new NameIdMapper.
```
xytan0056:
```suggestion
// NewNameIDMapper creates a new NameIDMapper.
```

internal/idmapper/idmapper.go
justinwon777: Yeah it'd be less wordy. Updated

xytan0056 (approved):

PR #179, opened 2026-07-10
refactor(orchestrator): add ResultToGetTargetGraphResponse and migrate call sites

justinwon777:
Adds orchestrator.ResultToGetTargetGraphResponse, copied from core/common. This is only called by orchestrator and the benchmark tool, so I migrated them here also. A follow-up PR will remove the duplicated code once all of core/common is migrated.

## Test Plan

unit test

### DIFF
--- a/orchestrator/targetgraph.go
+++ b/orchestrator/targetgraph.go
@@ -1,0 +1,174 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package orchestrator
+
+import (
+	"context"
+	"encoding/hex"
+
+	buildpb "github.com/bazelbuild/buildtools/build_proto"
+	"github.com/uber/tango/core/common"
+	"github.com/uber/tango/core/targethasher"
+	"github.com/uber/tango/tangopb"
+)
+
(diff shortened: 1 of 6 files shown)


orchestrator/targetgraph.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/uber/tango/tangopb"
+)
+
+// _cancelCheckInterval is how often we poll ctx.Err() inside per-target hot loops.
+// Picked to keep overhead negligible while still surfacing cancellation in <100ms
+// for typical target rates.
+const _cancelCheckInterval = 4096
+
+// ResultToGetTargetGraphResponse converts a Result to a GetTargetGraphResponse
+func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Result) ([]*tangopb.GetTargetGraphResponse, error) {
```
liujsj, line 33: Why is this part of orchestrator package? Given that it is trying to map the tango result to the proto response, it should be in own mapper package (not orchestrator) which is added as a dependency in the handler code (In this case controller since controller is acting as both handler + controller)

sywhang, line 33: we're leaking the proto definition to orchestrator now?

justinwon777:
> Why is this part of orchestrator package? Given that it is trying to map the tango result to the proto response, it should be in own mapper package (not orchestrator) which is added as a dependency in the handler code (In this case controller since controller is acting as both handler + controller)

> we're leaking the proto definition to orchestrator now?

Addressing both @liujsj @sywhang:
The way this is used right now is in orchestrator's `GetTargetGraph`:
```
responses, err := common.ResultToGetTargetGraphResponse(ctx, result)
...
err = storage.WriteGraphStream(ctx, b.storage, treehashPath, responses)
```
The proto response is not Tango's API response, but it stays in `GetTargetGraph` and used as the input to a storage API. So orchestrator package already depends on proto from other things and this PR isn't adding it. If we wanted to remove the proto dependency we need to have entities defined so orchestrator (and its deps) only works with those types, and the mapping logic moved to mapper (this function combines processing targethasher result to optimized targets then turning that into proto response). We already have work items that will cover those things, so my intention here is just moving the function out of common package to a more appropriate package, then the fixes come later. If you guys would prefer fixing it first, then moving it, we can do that too.

justinwon777:
discussed with team offline and we are going to make fixes first before trying to move it out of common

PR #180, opened 2026-07-10
[config] Tango configuration documentation

yushan8:
## Summary
Documents all user-facing configuration fields across service, repository, and storage sections.

## Test plan
- [x] `make build`
- [x] `make test`

## Stack
1. @ #180
2. #190
3. #191


docs/config/config.md
```diff
@@ -0,0 +1,65 @@
...
+| `streaming.max_num_metadata_entries` | No | `int` | `50000` | Max entries in the `Metadata` per stream message. |
+
+## `repository`
+
+A list of repository entries. Each entry configures how Tango analyzes a specific repository.
+
+| Field | Required | Type | Default | Description |
+|---|---|---|---|---|
+| `remote` | **Yes** | `string` | | The URL used to `git clone` the repository. Must be unique across all entries. Must match exactly what clients send in `BuildDescription.remote`. |
+| `bzlmod_enabled` | No | `bool` | `false` | Use Bzlmod for external dependency management. When disabled, `//external:all-targets` is added to Bazel queries for WORKSPACE-based repositories. |
```
liujsj: BzlMod is enforced in Bazel 9 and above. Why is the default false and not true given in a year that this is the standard.

docs/config/config.md
```diff
@@ -0,0 +1,65 @@
...
+| `bazel_command` | No | `string` | `""` | Override the Bazel binary path. Empty uses `bazel` from `$PATH`. |
+| `bazel_extra_args` | No | `[]string` | `[]` | Extra arguments passed to every Bazel invocation. |
+| `query_timeout_seconds` | No | `int64` | `600` | Bazel query timeout in seconds. |
+| `hash_external_repo_files` | No | `[]string` | `[]` | External repositories to hash per-file instead of using the default repo-level hash. By default, all files in an external repo share a single hash, so a version bump marks every target as changed. Listing a repo here hashes each file individually for finer change detection at the cost of slower computation. |
+| `excluded_files` | No | `[]string` | `[]` | Regex patterns for files to exclude from analysis (e.g. `"^@@?bazel_tools/"`). |
```
liujsj: What is "analysis" in this case? Should we be more descriptive here?

docs/config/config.md
```diff
@@ -0,0 +1,65 @@
...
+  #   max_num_targets: 250
+  #   max_num_changed_targets: 125
+  #   max_num_metadata_entries: 50000
+
+repository:
+  - remote: "https://github.com/uber/tango.git"
+    # excluded_files:
+    #   - "^@@?bazel_tools/"
+    # bzlmod_enabled: true
+    # query_timeout_seconds: 600
```
liujsj: Should the example be moved as part of where we are defining each configuration?

docs/config/config.md
```diff
@@ -0,0 +1,65 @@
...
+
+Controls worker pools, filesystem paths, and gRPC streaming behavior.
+
+| Field | Required | Type | Default | Description |
+|---|---|---|---|---|
+| `max_worker_pool_size` | **Yes** | `int` | | Number of concurrent worker workspaces per repository. Must be greater than 0. |
+| `workspaces_root` | **Yes** | `string` | | Root directory for all Tango on-disk state. Should be set to a persistent location in production. Tango creates the following structure under this path: `<workspaces_root>/<repo>/` for origin clones and `<workspaces_root>/.workers/<repo>/worker-{1..N}/` for worker checkouts. |
+| `streaming.max_num_targets` | No | `int` | `250` | Max `OptimizedTarget` entries per gRPC stream message. |
+| `streaming.max_num_changed_targets` | No | `int` | `125` | Max `ChangedTarget` entries per stream message. |
+| `streaming.max_num_metadata_entries` | No | `int` | `50000` | Max entries in the `Metadata` per stream message. |
```
liujsj: What does "OptimizedTarget", "ChangedTarget, "Metadata" refer to?

docs/config/config.md
```diff
@@ -0,0 +1,65 @@
...
+
+---
+
+## `service`
+
+Controls worker pools, filesystem paths, and gRPC streaming behavior.
+
+| Field | Required | Type | Default | Description |
+|---|---|---|---|---|
+| `max_worker_pool_size` | **Yes** | `int` | | Number of concurrent worker workspaces per repository. Must be greater than 0. |
```
liujsj: Doesn't this control the number of concurrent request that can be handled? Reading the current description, I am not sure what value I should be putting here

docs/config/config.md
yushan8: Good point - updated.

docs/config/config.md
yushan8: Added the reference to the proto file definition.

docs/config/config.md
yushan8: Yup, it also refers to the number of local clones. Updated to be more clear.

config/README.md
```diff
@@ -0,0 +1,62 @@
...
+| `streaming.max_num_targets` | No | `int` | `250` | Max number of target entries per gRPC stream message (see `OptimizedTarget` in `proto/tango.proto`). Keep messages under gRPC's default 4MB receive limit — raise this only if the client is configured to accept larger messages and you want fewer round-trips. |
+| `streaming.max_num_changed_targets` | No | `int` | `125` | Max number of changed target entries per stream message (see `ChangedTarget` in `proto/tango.proto`). Same tradeoff as `max_num_targets`. Default is lower because each entry carries both old and new targets (~2x the size). |
+| `streaming.max_num_metadata_entries` | No | `int` | `50000` | Max number of ID-to-name mapping entries per stream message (see `Metadata` in `proto/tango.proto`). Large monorepos can have hundreds of thousands of metadata entries — lowering this value splits them across more messages, reducing peak memory per message. |
```
liujsj: Open Question: Should this be a config? Without it, the service would still be able to provide a response (just more network calls).

The reason I bring this up is that all 3 config impact the response size. This is only needed if a client can handle larger response but in this case it is not obvious which one they should increase.

config/README.md
```diff
@@ -0,0 +1,62 @@
+# Tango Configuration
+
+Tango is configured via a single YAML file. Pass the file path to `config.Parse` at startup.
```
sbalabanov, line 3: plz indicate whether a service can pickup configuration changes without restart or not. Explain potential mechanisms for configuration delivery and app lifecycle management when config is changing.

sbalabanov (changes requested):

config/README.md
yushan8: I think the risk of using the defaults is if a monorepo has unusually large targets, long labels, big attribute blobs, long dep lists, it may reach the gRPC message limit even with default max sizes. I think the option is just nice to have so consumers who want to extend on the service can adjust the values to their needs.

config/README.md
yushan8: Updated

config/README.md
liujsj: The default are meant to protect this though so the monorepo has large targets shouldn't be a risk. 

I guess it isn't obvious to me on how user should adjust which value since in the end, they will get the total response so changing max_num_targets vs max_num_changed_targets vs max_num_metadata_entries does not matter. Given that, would it be better to somehow consolidate into one user config if you think that this should still be configurable.

config/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## `repository`
+
+A list of repository entries. Each entry tells Tango how to clone, and query for a specific repository — including which Bazel settings to use.
+
+```yaml
+repository:
+  - remote: "https://github.com/uber/tango.git"
+    bzlmod_enabled: true
+    # query_timeout_seconds: 600
```
sywhang: is this meant to be commented out? It is documented in the table below

config/README.md
yushan8: That's fair. Updated to be configurable with `max_message_bytes`

config/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+| Field | Required | Type | Default | Description |
+|---|---|---|---|---|
+| `remote` | **Yes** | `string` | | The URL used to `git clone` the repository. Tango clones from this URL and uses it as the lookup key for per-repo settings. Must be unique across all entries and match exactly what clients send in `BuildDescription.remote`. |
+| `bzlmod_enabled` | No | `bool` | `true` | Whether this repository uses Bzlmod for external dependency management. Bzlmod is enforced in Bazel 9+. Set to false only for repositories still using WORKSPACE — when disabled, `//external:all-targets` is added to Bazel queries. Setting this incorrectly causes query failures — it must match the repository's actual dependency management setup. |
```
sywhang, line 49: Is this right? i don't see default true being set anywhere in the codebase. Where is that set?

config/README.md
```diff
@@ -0,0 +1,68 @@
...
+| `bazel_command` | No | `string` | `""` | Override the Bazel binary path. When empty, Tango automatically downloads and caches [Bazelisk](https://github.com/bazelbuild/bazelisk) from GitHub. This lets Tango run without Bazel pre-installed and allows each repository to control its own Bazel version via `.bazelversion`. Set this when the repository requires a custom Bazel wrapper script. |
```
sywhang: usually configs like these have `path` suffix.

config/README.md
yushan8: Commented out means that it's optional, but I can uncomment it if it's not clear.

config/README.md
yushan8: This will be updated in subsequent PRs. This README.md has changes so we can finalize how the config will look. Will update now that the config fields is finalized.

config/README.md
yushan8: gotcha 👍

config/README.md
yushan8: https://github.com/uber/tango/pull/190/changes#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R84

PR #189, opened 2026-07-12
feat(entity): add domain types for BuildDescription and GetTargetGraphRequest

yushan8:
## Summary
- Adds `entity.BuildDescription`, `entity.ChangeRequest`, `entity.ComputationStrategy`, and `entity.GetTargetGraphRequest` as domain types, plus field-level doc comments.
- Adds mappers for`internal/mapper` proto->entity conversions to be used for mapping proto definitions to entities.
- These entities need to live outside of `internal` because `RepoManager` and `Orchestrator` interfaces accept these entities. Other consumers of tango may have their own implementations of these interfaces

Base for #187 and #192 , opened for review/discussion before merging.

## Test plan
- [x] `go build ./...`
- [x] `go test ./entity/... ./internal/mapper/...`

**Stack** (bottom to top):
1. @ #189
1. #193
1. #192
1. #187


internal/mapper/build_description.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/uber/tango/tangopb"
+)
+
+// ProtoToBuildDescription converts a proto BuildDescription to the domain type.
+func ProtoToBuildDescription(desc *tangopb.BuildDescription) entity.BuildDescription {
+	if desc == nil {
+		return entity.BuildDescription{}
+	}
+	return entity.BuildDescription{
+		Remote:         desc.GetRemote(),
```
justinwon777, line 25: should you validate empty remote and base sha here?

xytan0056: this is guarded before conversion in
[validateGetChangedTargetsRequest](https://github.com/uber/tango/blob/main/controller/getchangedtargets.go#L1041)

yushan8: Was checking how we do it internally, it seems like it's good practice to check for nil/empty values in the mapper functions. We can delete validateGetChangedTargetsRequest then and I'll add TODO.

justinwon777: yeah I saw that too and I think it makes sense cause all requests need to be mapped to some entity so the mapper does validation also.

xytan0056 (approved):

entity/computation_strategy.go
```diff
@@ -0,0 +1,33 @@
...
+// ComputationStrategy controls which graph computation method to use.
+type ComputationStrategy int
+
+const (
+	// ComputationStrategyInvalid is the zero value for an unrecognized or
+	// out-of-range strategy; never set intentionally.
+	ComputationStrategyInvalid ComputationStrategy = iota
+	// ComputationStrategyUnset means no strategy was specified by the caller;
+	// the orchestrator picks its own default.
+	ComputationStrategyUnset
```
xytan0056: nit: unset should be the first zero value, cuz that's technically "unset"

yushan8: Good catch, swapped the order so `ComputationStrategyUnset` is the zero value: 4bac9a6.

PR #190, opened 2026-07-12
config: update fields, defaults, and required checks to match README

yushan8:
## Summary
Aligns config field names, defaults, and required checks with config/README.md from #180.

- Renamed fields: `worker_pool_size` → `max_worker_pool_size`, `repo_manager_clone_path` → `workspaces_root`, `chunking` → `streaming` (with sub-fields renamed to `max_num_targets`/`max_num_changed_targets`/`max_num_metadata_entries`), `query_timeout` → `query_timeout_seconds`.
- `config.Parse` now enforces `service.workspaces_root` as required and applies documented defaults: `bzlmod_enabled` defaults to `true`, `query_timeout_seconds` defaults to `600`, streaming fields default to `250`/`125`/`50000`.
- Added Go doc comments for config fields documented in the README.
- Updated all callsites and YAML fixtures for the renames.
- Added `config/config_test.go` covering new validation and defaults.

## Test plan
- [x] `make build`
- [x] `make test`

## Stack
1. #180
2. @ #190
3. #191

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,18 +76,14 @@ func ParseBytes(yamlBytes []byte) (*Config, error) {
 	}
 
 	// --- service validation and defaults ---
-	if config.Service.WorkerRootPath != "" && config.Service.RepoManagerClonePath == "" {
-		return nil, fmt.Errorf("service.repo_manager_clone_path must be set when worker_root_path is specified")
-	}
-	// Default: os.TempDir()/tango-repo-manager.
-	if config.Service.RepoManagerClonePath == "" {
-		config.Service.RepoManagerClonePath = filepath.Join(os.TempDir(), "tango-repo-manager")
+	if config.Service.WorkspacesRootPath == "" {
+		return nil, fmt.Errorf("service.workspaces_root_path must be set")
 	}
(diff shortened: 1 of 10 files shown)


config/config.go
```diff
@@ -22,6 +22,17 @@ import (
...
+const (
+	_defaultBazelQueryTimeoutSeconds = 600
+	_defaultMaxNumTargets            = 250
+	_defaultMaxNumChangedTargets     = 125
+	_defaultMaxNumMetadataEntries    = 50_000
+)
+
+var (
+	_bzlmodEnabledDefault = true
+)
```
liujsj: Why is this not in the const block too?

liujsj (approved):

config/config.go
yushan8: It's because the value is `*bool` not `bool`. Since the default is now true we need to make sure the value was unset. If we made it a `bool`, the default is false. There is no way to tell if it's false because it's unset or false because the user set it. So to check if bzlmod was disabled, we would check `BzlmodEnabled != nil && *BzlmodEnabled == false`.

xytan0056 (approved):

PR #191, opened 2026-07-13
config: remove fields not documented in config/README.md

yushan8:
## Summary
Removes config fields not documented in config/README.md so the Go structs match the documented surface exactly.

- Removed `ServiceConfig.WorkerRootPath`.
- `example/main.go` derives the worker checkout path as `workspaces_root/.workers` instead of reading a separate config field.
- Updated tests and YAML fixtures.

## Test plan
- [x] `make build`
- [x] `make test`

## Stack
1. #180
2. #190
3. @ #191

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"fmt"
 	"os"
-	"path/filepath"
 
 	yaml "github.com/goccy/go-yaml"
 )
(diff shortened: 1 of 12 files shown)


config/config.go
```diff
@@ -72,12 +72,10 @@ func Parse(configFilePath string) (*Config, error) {
 	if config.Service.WorkspacesRoot == "" {
 		return nil, fmt.Errorf("service.workspaces_root must be set")
 	}
-	if config.Service.WorkerRootPath == "" {
-		config.Service.WorkerRootPath = filepath.Join(config.Service.WorkspacesRoot, ".workers")
-	}
 	if config.Service.MaxWorkerPoolSize <= 0 {
 		return nil, fmt.Errorf("service.max_worker_pool_size must be > 0, got %d", config.Service.MaxWorkerPoolSize)
 	}
+	config.Service.WorkerRootPath = filepath.Join(config.Service.WorkspacesRoot, ".workers")
```
liujsj: Should we keep this outside of config.go as it is not a field user can provide

yushan8: Yup good point.

xytan0056 (approved):

PR #193, opened 2026-07-13
refactor: Update callsites for cachekey methods and mapper.ProtoToBuildDescription

yushan8:
## Summary
- Update references to cache-path methods to `internal/cachekey`
- Callers (`controller`, `orchestrator`) still only hold proto `BuildDescription` values at this point, so each converts once, inline, via the existing `mapper.ProtoToBuildDescription` before calling the new helpers.

## Test plan
- [x] `go build ./...`
- [x] `go test ./controller/... ./orchestrator/... ./internal/mapper/... ./internal/cachekey/... ./core/common/...`
- [x] `bazel build`/`bazel test` on the same packages

**Stack** (bottom to top):
1. #189
1. @ #193
1. #192
1. #187

### DIFF
--- a/controller/BUILD.bazel
+++ b/controller/BUILD.bazel
@@ -17,6 +17,8 @@ go_library(
         "//config",
         "//core/common",
         "//core/storage",
+        "//internal/cachekey",
+        "//internal/mapper",
         "//orchestrator",
         "//tangopb",
         "@com_github_uber_go_tally//:tally",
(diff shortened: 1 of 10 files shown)


justinwon777:
I had a [PR](https://github.com/uber/tango/pull/177) to create a cachekey package with these functions. Do you think these belong in mapper instead?

xytan0056 (approved):

PR #194, opened 2026-07-13
log errors once in each RPC

justinwon777:
## Why?

Currently logger.Error calls are made at the error return site, which risks double logging and also means the logged error is not the wrapped error that is returned.
## What?

Make sure each RPC handler has only one logger.Error call which happens in defer.
## Test Plan

Logging only change

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -43,8 +43,13 @@ type job struct {
 func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, stream pb.TangoServiceGetChangedTargetsYARPCServer) (retErr error) {
 	scope := c.scope.SubScope("get_changed_targets")
 	scope.Counter("calls").Inc(1)
+	logger := c.logger.WithLazy(
+		zap.Any("first_revision", request.GetFirstRevision()),
+		zap.Any("second_revision", request.GetSecondRevision()),
+	)
 	defer func() {
 		if retErr != nil {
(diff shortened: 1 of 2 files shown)


controller/getchangedtargets.go
```diff
@@ -43,8 +43,13 @@ type job struct {
 func (c *controller) GetChangedTargets(request *pb.GetChangedTargetsRequest, stream pb.TangoServiceGetChangedTargetsYARPCServer) (retErr error) {
 	scope := c.scope.SubScope("get_changed_targets")
 	scope.Counter("calls").Inc(1)
+	logger := c.logger.With(
+		zap.Any("first_revision", request.GetFirstRevision()),
+		zap.Any("second_revision", request.GetSecondRevision()),
+	)
```
sywhang: use LazyWith here so that we're not allocating these on hot path

controller/gettargetgraph.go
```diff
@@ -33,8 +33,12 @@ import (
 func (c *controller) GetTargetGraph(request *pb.GetTargetGraphRequest, stream pb.TangoServiceGetTargetGraphYARPCServer) (retErr error) {
 	scope := c.scope.SubScope("get_target_graph")
 	scope.Counter("calls").Inc(1)
+	logger := c.logger.With(
```
sywhang: same thing here

sywhang (approved):
please address the two comments before merging.

controller/getchangedtargets.go
justinwon777: updated

PR #195, opened 2026-07-14
replace NotFoundError struct with ErrNotFound sentinel

justinwon777:
## Why?

NotFoundError is used to check cache and the Path value is never used by callers, so there is no need for this error to be a struct.
## What?

Change the error to a sentinel
## Test Plan

unit test

### DIFF
--- a/core/storage/storage.go
+++ b/core/storage/storage.go
@@ -16,22 +16,22 @@ package storage
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"io"
 )
 
-// NotFoundError represents an error when a blob is not found in the storage.
-type NotFoundError struct {
-	Path string
-}
+// ErrNotFound is returned when a blob is not found in the storage.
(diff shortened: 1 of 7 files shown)


core/storage/storage.go
```diff
@@ -16,22 +16,22 @@ package storage
...
+	return errors.Is(err, ErrNotFound)
 }
 
-// IsNotFound checks if an error is a NotFoundError
-func IsNotFound(err error) bool {
-	_, ok := err.(*NotFoundError)
-	return ok
+// NewNotFoundError wraps ErrNotFound with the key that was not found.
+func NewNotFoundError(key string) error {
+	return fmt.Errorf("get %q: %w", key, ErrNotFound)
```
sbalabanov: string message is the only thing that the inspector sees (there is no stack).
please make sure the error message says this is a storage error.

sbalabanov (approved):

sywhang (approved):

PR #197, opened 2026-07-14
[PENDING] refactor(graphrunner): route metrics through observability/metrics

xytan0056:
## [PENDING] — do not merge until pr6-orchestrator lands

Routes the native graphrunner's metrics through `observability/metrics`.

- The graphrunner **reads the request emitter from `ctx` via `metrics.FromContext`** (the orchestrator forwards the same ctx the controller seeded), so it no longer takes a `tally.Scope` at construction.
- Substep timers (`bazel_query_duration`, `git_file_hashes_duration`, `target_hash_duration`) and the target-count gauge move to shared metric-name constants under the `graph_runner` op sub-scope. `target_count` → `targets_count`, still a gauge to preserve the time-series shape.
- With the runner taking its emitter from ctx, the orchestrator's `Scope` plumbing (`Params.Scope` + the threaded field) is now dead and removed.

### Stacking / dependency

Based on **pr6-orchestrator**. Independent of #174. Same as pr6: the emitter is seeded by the controller PR (pr5); until then `FromContext` returns the noop emitter (safe).

### DIFF
--- a/graphrunner/BUILD.bazel
+++ b/graphrunner/BUILD.bazel
@@ -15,7 +15,7 @@ go_library(
         "//core/git",
         "//core/targethasher",
         "//core/workspace",
-        "@com_github_uber_go_tally//:tally",
+        "//observability/metrics",
     ],
 )
 
--- a/graphrunner/native.go
+++ b/graphrunner/native.go
@@ -18,12 +18,12 @@ import (
 	"context"
 	"time"
 
-	"github.com/uber-go/tally"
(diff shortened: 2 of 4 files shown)


xytan0056:
Superseded by #198. GitHub auto-marked this PR as "merged" when the stack order was flipped so the graphrunner change lands before the orchestrator change (it couldn't be reopened). The graphrunner work now lives in #198 (based on pr4-context).

PR #203, opened 2026-07-14
config: bound stream chunks by real message size, not estimated

yushan8:
## Summary
Migrates all callsites to the new `internal/mapper` chunking primitives from #206, consolidates config, removes old code, and replaces proto-typed storage interfaces with entity-typed ones.

- Replaces `service.streaming.{max_num_targets,max_num_changed_targets,max_num_metadata_entries}` with a single `service.max_message_bytes` byte budget, matching `config/README.md`.
- Introduces `mapper.GraphChunkReader` (wraps `storage.GraphReader`, converts proto → `entity.GraphChunk` on read) and `mapper.WriteTargetGraph` (converts `entity.TargetGraph` → proto, writes to storage).
- `Orchestrator.GetTargetGraph` now returns `*mapper.GraphChunkReader` instead of `storage.GraphReader` — the orchestrator no longer touches proto types.
- Controller reads `entity.GraphChunk` from the reader and converts to proto at stream-send time via `mapper.GraphChunkToProto`.
- `core/common/utils.go` stripped to just `ToShortRemote` — all chunking code, chunk constants, and proto imports removed.
- `config.Parse()` defaults `MaxMessageBytes` to ~4.25 MB when unset.

## Test plan
- [x] `make build`
- [x] `make test` — 19/19 pass
- [x] `make gazelle && gofmt -w . && goimports -w .`

## Stack
1. #180
2. #190
3. #191
4. #206
5. @ #203


internal/mapper/chunk.go
```diff
@@ -0,0 +1,139 @@
...
+		currentBytes += itemBytes
+	}
+	return append(chunks, current)
+}
+
+// chunkTargets splits OptimizedTarget entries into GetTargetGraphResponse
+// messages, each bounded by maxMessageBytes of real wire size.
+func chunkTargets(targets []*tangopb.OptimizedTarget, maxMessageBytes int) []*tangopb.GetTargetGraphResponse {
+	chunks := BySize(targets, maxMessageBytes)
+	responses := make([]*tangopb.GetTargetGraphResponse, 0, len(chunks))
```
yushan8: Moved chunkTargets to internal/mapper to avoid leaking proto definitions into other packages. We also want to utilize proto's built-in `Size()` to calculate the size of the message we're sending over the stream.

orchestrator/native_orchestrator.go
```diff
@@ -208,10 +209,7 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, req entity.GetT
 	if err != nil {
 		return nil, fmt.Errorf("compute target graph: %w", err)
 	}
-	responses, err := common.ResultToGetTargetGraphResponse(ctx, result,
-		b.config.Service.Streaming.MaxNumTargets,
-		b.config.Service.Streaming.MaxNumMetadataEntries,
-	)
+	responses, err := mapper.ResultToGetTargetGraphResponse(ctx, result, b.config.Service.MaxMessageBytes)
```
justinwon777: this is bringing in proto to orchestrator which is what we're trying to prevent? Can `storage.WriteGraphStream` input not be the proto response so we can get rid of having to map result to proto here?

yushan8: Good idea, going to close this and refactor that out here https://github.com/uber/tango/pull/212

PR #204, opened 2026-07-14
test(integration): add GetChangedTargets benchmarks

yushan8:
## Summary
- Adds a `make bench` target that measures `GetChangedTargets` latency with two benchmarks:
  - **Cold** — 15 iterations over unique commit pairs, each a cache miss exercising the full pipeline (git checkout, bazel query, compare, stream)
  - **Cached** — 15 iterations hitting the in-memory cache, measuring streaming overhead only
- Adds `startServerWithLogger` to allow benchmarks to inject a nop logger, suppressing service log noise
- Measurement only — no pass/fail thresholds, not part of `make test` / `make test-integration`, and never invoked from CI
- Widens `repoRemote`/`startServer`/`newClient`/`getChangedTargets` in `integration_test.go` from `*testing.T` to `testing.TB` so the existing integration harness can be reused from `Benchmark*` functions

## Test plan
- [x] `make gazelle` — `integration/BUILD.bazel` picked up the new `benchmark_test.go`
- [x] `make test-integration` — existing integration tests still pass unchanged
- [x] `make bench` — both benchmarks run 15 iterations and report `ns/op`, exits 0


Makefile
```diff
@@ -90,6 +104,7 @@ help:
 	@echo "  make build            - Build all targets"
 	@echo "  make test             - Run all tests"
 	@echo "  make test-integration - Run integration tests (slow)"
+	@echo "  make benchmark        - Run GetChangedTargets benchmarks (measurement only, not in CI)"
```
sywhang:
```suggestion
	@echo "  make bench           - Run all benchmark tests
```

sywhang (commented):
What's the current measured benchmark on this on your local box?

yushan8:
> What's the current measured benchmark on this on your local box?
```
BenchmarkGetChangedTargets_SmallDiff-10       3   503320 ns/op   0.0005030 avg_cached_call_sec/op   44.90 first_uncached_call_sec
BenchmarkGetChangedTargets_MediumDiff-10      3   466181 ns/op   0.0004658 avg_cached_call_sec/op   41.07 first_uncached_call_sec
BenchmarkGetChangedTargets_LargeDiff-10       3  1451639 ns/op   0.001451 avg_cached_call_sec/op    38.21 first_uncached_call_sec
```
uncached run is the initial request, cached run is the subsequent calls with the same request.

Makefile
```diff
@@ -21,6 +21,24 @@ test-integration:
...
 	@echo "Integration tests passed!"
 
+# Run GetChangedTargets benchmarks against fixed, checked-in commit pairs.
+# Measurement only: not part of `make test` / `make test-integration` and not
+# run in CI, so a slow benchmark never fails the build.
+bench:
+	@echo "Running benchmarks..."
+	@$(BAZEL) test //integration:integration_test \
+		--test_output=all \
+		--test_arg=-test.run=^$$ \
```
sywhang, line 31: let's run it with higher iteration by default. like 15 or 20.

yushan8: I decided changed the test to run `getChangedTargets` on different commits to test cold cache calls. Since each iteration needs a unique commit pair for a cache miss, b.N is capped at the number of pairs we have. `BenchmarkGetChangedTargets_Cold` tests for the requests with cache miss (full bazel query) with 15 distinctive commit pairs, while the other `BenchmarkGetChangedTargets_Cached` tests for the duration of cache hits.
This lets us measure the endpoint latency without the server startup.

yushan8: Updated benchtime to run 15x

sywhang (approved):

PR #206, opened 2026-07-15
feat(mapper): add wire-size-based chunking and ResultToGetTargetGraphResponse

yushan8:
## Summary
Introduces wire-size-based chunking and a proto-free domain type for target graphs, separating entity conversion from proto serialization.

- Add `entity.OptimizedTarget` and `entity.TargetGraph` as proto-free domain types for target graph data.
- Add `ResultToTargetGraph`: converts `targethasher.Result` → `entity.TargetGraph` (proto-free, suitable for orchestrator use).
- Add `TargetGraphToProto`: converts `entity.TargetGraph` → chunked `[]*tangopb.GetTargetGraphResponse` (proto serialization + ID mapping + chunking).
- Add `internal/mapper/chunk.go` with `BySize` (generic proto chunker bounded by real serialized size via `Size()`) and `ChunkMetadata` (map splitter by measured entry wire size).
- `BySize` always returns at least one chunk (empty input → single empty chunk); non-first empty chunks return an error.
- Both old (`core/common`) and new (`internal/mapper`) implementations coexist — callsite migration follows in #203.

This split enables a future PR to have the orchestrator work with `entity.TargetGraph` directly, removing proto from the orchestrator layer entirely.

## Test plan
- [x] `make build`
- [x] `make test` — `ResultToTargetGraph` (entity conversion), `TargetGraphToProto` (empty graph), `ResultToGetTargetGraphResponse` (chunking at various sizes), `BySize` / `ChunkMetadata` unit tests
- [x] `make gazelle && gofmt -w . && goimports -w .`

## Stack
1. #180
2. #190
3. #191
4. @ #206
5. #203


internal/mapper/target_graph.go
```diff
@@ -24,3 +31,109 @@ func ProtoToGetTargetGraphRequest(req *tangopb.GetTargetGraphRequest) (entity.Ge
 		BypassCache:       req.GetBypassCache(),
 	}, nil
 }
+
+// ResultToGetTargetGraphResponse converts a targethasher.Result into chunked
+// GetTargetGraphResponse messages ready for streaming or storage. Each message
+// is bounded by maxMessageBytes of real wire size (see chunkTargets).
+func ResultToGetTargetGraphResponse(ctx context.Context, result targethasher.Result, maxMessageBytes int) ([]*tangopb.GetTargetGraphResponse, error) {
```
justinwon777: I feel like this function is doing a lot of logic that maybe shouldn't be in mapper. This could be broken into two. First is taking targethasher.Result and processing it into optimized targets. This part can stay in orchestrator (optimized target needs an entity). Second is turning optimized targets into the proto response, so the mapper can do just this part.

yushan8: Yes good point. Going to close this in favor of https://github.com/uber/tango/pull/211 to split it up better.

PR #207, opened 2026-07-16
feat: add internal/url package for URL helpers

justinwon777:
Create `internal/url` package which contains URL helper functions. Add `ToShortRemote` which is currently in `core/common` and `GetReqURLsHash` from `core/cachekey`. Follow up will migrate usages to the new package.

## Test Plan

unit test


sywhang (commented):
Does this make more sense to place under core/git

justinwon777:
> Does this make more sense to place under core/git

core/git is for the git interface which this has nothing to do with so I felt like this didn't belong there, or do you mean making the package core/git/utils? That would make more sense to me.

justinwon777:
Changed it to `core/git/utils` so outside packages can import it

xytan0056 (approved):

xytan0056:
remove this too?
https://github.com/uber/tango/blob/37de82b17406fe79a450c0a653a41956f43aa835/core/repomanager/repo_manager.go#L217

justinwon777:
> remove this too?
> 
> https://github.com/uber/tango/blob/37de82b17406fe79a450c0a653a41956f43aa835/core/repomanager/repo_manager.go#L217

Will do this in migration PR after this

justinwon777:
changed it to `internal/url` package to make this an intentional package for url helpers instead of general utils.

internal/url/url_test.go
```diff
@@ -0,0 +1,98 @@
...
+			want: md5hex("a", "b"),
+		},
+		{
+			name: "multiple sorted",
+			in:   []entity.ChangeRequest{{URL: "b"}, {URL: "a"}},
+			want: md5hex("a", "b"),
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
sywhang: these are no longer needed after loopvar change in Go 1.22. Remove.

sywhang (approved):

PR #208, opened 2026-07-16
refactor(repomanager): classify ErrPoolTimeout as ErrorInfraRetryable

yushan8:
## Summary
- Expose `repomanager.ErrPoolTimeout` as a sentinel error so upper layers can classify it.
- Pool timeout should be classified as a `RetryableInfra` error since requests are timing out waiting to Lease a workspace (only on context.DeadlineExceeded).
- Add `classifyLeaseError` in the orchestrator to map `ErrPoolTimeout` → `ErrorInfraRetryable` and other errors → `ErrorInfra`
- Add tests for sentinel error propagation and classification at both layers

## Test plan
- [x] `bazel test //core/repomanager:repomanager_test` — verifies `ErrPoolTimeout` sentinel is in the error chain on pool timeout
- [x] `bazel test //orchestrator:orchestrator_test` — verifies `classifyLeaseError` maps sentinels correctly, and integration tests confirm classification through `GetTargetGraph`


core/repomanager/repo_manager.go
```diff
@@ -145,7 +153,11 @@ func (r *repoManager) Lease(ctx context.Context, desc entity.BuildDescription) (
 	select {
 	case slot = <-pool.avail:
 	case <-ctx.Done():
-		return nil, ctx.Err()
+		err := ctx.Err()
+		if errors.Is(err, context.DeadlineExceeded) {
+			return nil, fmt.Errorf("pool for repo %s: %w", repo, ErrPoolTimeout)
```
justinwon777: I think it'd be good to not lose the context error in the wrap. What about doing this?
```
return nil, fmt.Errorf("%w: %w", ErrPoolTimeout, err)
```

justinwon777 (approved):

PR #214, opened 2026-07-17
feat(observability/metrics): add Emitter and Begin/Complete lifecycle helpers

xytan0056:
## Summary

Adds the `observability/metrics` package per the merged design doc (`docs/observability/metrics.md`, #158): a thin, concrete wrapper over `tally.Scope` pinning the path shape `<scope>.<op>.<name>`.

- **`emitter.go`** — `Emitter` (`New`/`Nop`/`Tagged`/`Counter`/`DurationHistogram`/`ValueHistogram`) plus the `Begin`/`Complete` lifecycle helpers over an `Op` handle.
- **`names.go`** — shared op / tag-key / result-value constants and `Outcome(err)` (only `context.Canceled` → cancelled; `DeadlineExceeded` → failure, a genuine timeout).

## Test plan
CI


observability/metrics/names.go
```diff
@@ -0,0 +1,53 @@
...
+
+package metrics
+
+import (
+	"context"
+	"errors"
+)
+
+// Operation (op) names live in each consuming package's metrics.go, named after
+// the interface method they measure (e.g. "GetTargetGraph", "Compute").
```
yushan8, line 23: Did we want to include these op names in here according to the RFC? https://github.com/uber/tango/pull/158/changes#diff-fee8d84eb3fa824026b461d0d295f53ac0c736def0a6f1555f19ef563fa841dcR127

yushan8 (approved):

observability/metrics/names.go
xytan0056: I initially did it in a centralized place in names.go than I figure it's kinda weird to have to do metric.Opxxx
op names are package-specific constants, not shared vocabulary

PR #215, opened 2026-07-17
refactor(controller): route metrics through observability/metrics Emitter

xytan0056:
Ports the controller off raw `tally.Scope` counters/timers to `*metrics.Emitter` with the `Begin`/`Complete` lifecycle helper.

- `Params` keeps `Scope tally.Scope`; `NewController` subscopes it to `controller` and builds its own emitter (nil → no-op). Same for every component — no injected emitter.
- Each RPC bakes the `repo` tag once, then `Begin`/`Complete` emit `start` + result-tagged `finish` (replacing `calls`/`success`/`failure` + `total_duration` timers).
- Sub-ops `get_graph` and `compare_target_graphs` get their own lifecycle; `compare_target_graphs` flattens to its own op.
- Other timers → `DurationHistogram` with package-level buckets (`controller/metrics.go`); `emitFailureMetric` keeps the `failure_type`/`failure_reason` axis.

1. #214.


controller/metrics.go
```diff
@@ -0,0 +1,48 @@
...
+var (
+	// requestFinishBuckets covers whole-operation finish durations, from a fast
+	// cache hit (~1ms) to a full compute (minutes): exponential 1ms..~70m.
+	requestFinishBuckets = tally.MustMakeExponentialDurationBuckets(time.Millisecond, 4, 12)
+
+	// stepDurationBuckets covers sub-operation durations: cache reads, sends,
+	// graph fetches, and comparison phases. These can run long (a graph fetch
+	// or compare may take tens of minutes), so the range extends to ~80m:
+	// exponential 1ms..~80m.
+	stepDurationBuckets = tally.MustMakeExponentialDurationBuckets(time.Millisecond, 3, 15)
```
yushan8: 70-80 minutes sounds pretty extreme. Do need buckets that go that high? Maybe limit it to 30 minutes?

yushan8 (approved):

controller/metrics.go
xytan0056: a few extra buckets is no big deal.  we don't lose visibility into tail cases. plus the long tails are still recorded. the granularity is not affected

PR #216, opened 2026-07-17
refactor(orchestrator): route metrics through observability/metrics Emitter

xytan0056:
Ports `native_orchestrator`'s `GetTargetGraph` off raw `tally.Scope` to `*metrics.Emitter` with the `Begin`/`Complete` helper.

- `Params` keeps `Scope tally.Scope`; the orchestrator self-subscopes to `orchestrator` and builds its own emitter (still forwards the scope to the graph runner).
- `start` + result-tagged `finish` (replaces `calls`/`success`/`failure`).
- Per-step duration histograms under `get_target_graph`: `lease_duration`, `checkout_duration`, `apply_requests_duration`, `cache_read_duration`, `compute_duration`, `cache_write_duration`.

Stacked on #215.


orchestrator/native_orchestrator.go
```diff
@@ -125,9 +117,11 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, req entity.GetT
 	if !ok {
 		return nil, fmt.Errorf("no repository configuration found for remote %q", remote)
 	}
+	leaseStart := time.Now()
 	ws, err := b.repoManager.Lease(ctx, build)
```
xytan0056, line 126: TODO: emit metrics before Lease error

orchestrator/native_orchestrator.go
```diff
@@ -83,12 +87,15 @@ func NewNativeOrchestrator(appCtx context.Context, p Params) (Orchestrator, erro
 	if scope == nil {
 		scope = tally.NoopScope
 	}
+	scope = scope.SubScope("orchestrator")
+	emitter, _ := metrics.New(scope)
```
yushan8: Should we fallback on `metrics.Nop()`? Seems like we're always silently discarding the error. I think we should either return the error or fallback to `Nop()`.

orchestrator/native_orchestrator.go
```diff
@@ -99,23 +106,8 @@ func NewNativeOrchestrator(appCtx context.Context, p Params) (Orchestrator, erro
...
-			}
-			scope.Tagged(map[string]string{
-				"failure_type":   ce.Type(),
-				"failure_reason": ce.Reason(),
-			}).Counter("failure_type").Inc(1)
-		} else {
-			scope.Counter("success").Inc(1)
-		}
-	}()
+	op := metrics.Begin(b.emitter, opGetTargetGraph, stepDurationBuckets)
```
yushan8: Should we have the `repo` tagged in the emitter as well?

orchestrator/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+package orchestrator
+
+import (
+	"time"
+
+	"github.com/uber-go/tally"
+)
+
+// opGetTargetGraph is snake_cased after the Orchestrator.GetTargetGraph method.
+const opGetTargetGraph = "get_target_graph"
```
yushan8:
```suggestion
const _opGetTargetGraph = "get_target_graph"
```

orchestrator/metrics.go
```diff
@@ -0,0 +1,29 @@
...
+
+// stepDurationBuckets covers whole-operation and individual pipeline steps
+// (lease, checkout, apply, cache read/write, compute). A compute on a large
+// repo can run for hours, so the range extends to ~4h: exponential 1ms..~4h.
+var stepDurationBuckets = tally.MustMakeExponentialDurationBuckets(time.Millisecond, 3, 16)
```
yushan8:
```suggestion
var _stepDurationBuckets = tally.MustMakeExponentialDurationBuckets(time.Millisecond, 3, 16)
```

yushan8 (approved):

orchestrator/native_orchestrator.go
xytan0056: should already be tagged in controller. but good piont, just in case NewNativeOrchestrator is called separately

PR #222, opened 2026-07-20
refactor: replace proto serialization with JSON and migrate callsites

yushan8:
## Summary
Intent: Proto definitions are currently leaking into the storage. We should `Put` the actual entities to be encoded/decoded from the storage, not the proto definitions.
- Remove all proto imports from `core/storage/` — serialize entity structs directly via `encoding/json`
- Move write functions from reader files into `graphwriter.go`
- Controller reads `entity.GetTargetGraphResponse` and `entity.GetChangedTargetsResponse` internally, converting to proto via mapper only at `stream.Send`
- Changed-targets pipeline fully migrated to entity types
- Orchestrator uses `mapper.ResultToTargetGraph` + `SplitBySize`
- Consolidate `service.chunking.*` into a single `max_message_bytes` config field
- Remove old chunking code from `core/common`
- Update `query-bench` example

## Test plan
- [x] `grep -r "tangopb\|gogo/protobuf" core/storage/` returns zero results
- [x] `make build && make test` — all 19 test targets pass
- [x] `make gazelle` — BUILD files in sync

## Stack
1. #210 (entity types)
2. #211 (mappers + streaming)
3. **#222 — this PR** (storage JSON + callsite migration + cleanup)

Ran `integration tests` and `make bench` to make sure storage is compatible with current changes.


config/service_config.go
```diff
@@ -16,23 +16,12 @@ package config
...
-	// ChangedTargetChunkSize is the max number of ChangedTarget entries per stream message.
-	// ChangedTarget carries both old and new targets (~2× the size of a regular target).
-	ChangedTargetChunkSize int `yaml:"changed_target_chunk_size"`
-	// MetadataMapChunkSize is the max number of entries per metadata map chunk.
-	// Applies to target_id_mapping and attribute_string_value_mapping.
-	MetadataMapChunkSize int `yaml:"metadata_map_chunk_size"`
-}
+// defaultMaxMessageBytes is the fallback max serialized size per streamed
+// message (~4.25 MB), well under the 64 MB default gRPC limit.
+const defaultMaxMessageBytes = 4_250_000
```
xytan0056: can this be exported? Multiple palces are referencing it

core/storage/reader.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2026 Uber Technologies, Inc.
+// Copyright (c) 2025 Uber Technologies, Inc.
```
xytan0056: 2026

mapper/mapper.go
xytan0056, line 1: now we have internal/mapper and mapper. How about `optimized`?

controller/getchangedtargets.go
```diff
@@ -449,39 +448,36 @@ func (c *controller) compareTargetGraphs(ctx context.Context, scope tally.Scope,
...
-					ChangedTargets: changed[i:end],
-				},
-			},
-		})
+	// Entity ChangedTarget doesn't implement Sizer, so convert to proto for size measurement
+	// (same pattern as ChunkTargetGraph) and map groups back by index.
+	tempResp := mapper.ChangedTargetsResponseToProto(&entity.GetChangedTargetsResponse{ChangedTargets: changed})
+	protoChanged := tempResp.GetItem().(*pb.GetChangedTargetsResponse_ChangedTargets).ChangedTargets.GetChangedTargets()
+	var results []entity.GetChangedTargetsResponse
+	changedGroups, err := streaming.SplitBySize(protoChanged, c.maxMessageBytes)
```
xytan0056: where is streaming.SplitBySize defined?  I can't find it
looks like we called `ChangedTargetsResponseToProto` twice here , once for splitting, once for sending later

controller/getchangedtargets.go
```diff
@@ -449,39 +448,36 @@ func (c *controller) compareTargetGraphs(ctx context.Context, scope tally.Scope,
...
-		results = append(results, &pb.GetChangedTargetsResponse{
-			Item: &pb.GetChangedTargetsResponse_ChangedTargets{
-				ChangedTargets: &pb.ChangedTargets{},
-			},
+	idx := 0
+	for _, group := range changedGroups {
+		results = append(results, entity.GetChangedTargetsResponse{
+			ChangedTargets: changed[idx : idx+len(group)],
 		})
+		idx += len(group)
```
xytan0056: actually looks like the whole `ChangedTargetsResponseToProto` -> `protoChanged` => `SplitBySize` ->  `changedGroups` is just for len(group) ? seems too expensive, can we feed something to SplitBysize without the conversion?

xytan0056 (commented):
nits

example/cmd/query-bench/main.go
```diff
@@ -112,18 +116,20 @@ func run() error {
 		totalDuration += elapsed
 		fmt.Printf("run %d: targethasher: %v  (%d targets)\n", i+1, elapsed.Round(time.Millisecond), len(targethasherResult.TargetNames))
 		start = time.Now()
-		response, err := common.ResultToGetTargetGraphResponse(
```
xytan0056, line 115: reminder to remove dead code in common, now that a bunch of functions in common are not used

core/storage/changedtargetsreader.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2026 Uber Technologies, Inc.
+// Copyright (c) 2025 Uber Technologies, Inc.
```
xytan0056: 2026

core/storage/reader.go
```diff
@@ -16,55 +16,43 @@ package storage
...
 	resp, err := st.Get(ctx, DownloadRequest{Key: key})
 	if err != nil {
 		return nil, err
 	}
-	return &reader[T, PT]{
-		rc:      gogio.NewDelimitedReader(resp.ReadCloser, maxMessageSize),
-		isEmpty: isEmpty,
+	return &reader[T]{
+		rc:  resp.ReadCloser,
+		dec: json.NewDecoder(resp.ReadCloser),
```
xytan0056, line 56: could be backwards incompatible here, the old cache is still proto based. Would that be an accepted risk?

xytan0056 (approved):

controller/getchangedtargets.go
```diff
@@ -429,50 +428,48 @@ func (c *controller) compareTargetGraphs(ctx context.Context, e *metrics.Emitter
...
+	start2 := 0
+	currentBytes := 0
+	for i := range changed {
+		itemBytes := changed[i].Size()
+		if i > start2 && currentBytes+itemBytes > c.maxMessageBytes {
+			results = append(results, entity.GetChangedTargetsResponse{
+				ChangedTargets: changed[start2:i],
+			})
+			start2 = i
+			currentBytes = 0
```
liujsj: Isn't there an internal function in streaming to do this?

https://github.com/uber/tango/blob/main/internal/streaming/streaming.go#L21C6-L21C17

example/cmd/query-bench/main.go
yushan8: yes, common is completely removed in this PR.

controller/controller.go
```diff
@@ -94,12 +81,7 @@ func NewController(appCtx context.Context, p Params) pb.TangoYARPCServer {
 // The returned cancel function MUST be deferred; it releases the
 // context.AfterFunc handle so we do not leak a watcher past the request.
 func (c *controller) linkRequestCtx(reqCtx context.Context) (context.Context, context.CancelFunc) {
-	// Derive a per-request ctx whose cancel only affects this ctx and its
-	// children — it never propagates up to reqCtx.
 	ctx, cancel := context.WithCancel(reqCtx)
-	// Register a one-shot watcher that cancels the derived ctx if appCtx fires.
-	// AfterFunc only observes appCtx; it never cancels it. stop() deregisters
-	// the watcher so the closure is not retained past the request.
```
liujsj, line 98: Was there a reason to delete these comments? The code change looks like it has nothing to do with this but got deleted

core/storage/reader.go
```diff
@@ -16,55 +16,43 @@ package storage
...
+func newReader[T any](ctx context.Context, st Storage, key string) (*reader[T], error) {
 	resp, err := st.Get(ctx, DownloadRequest{Key: key})
 	if err != nil {
 		return nil, err
 	}
-	return &reader[T, PT]{
-		rc:      gogio.NewDelimitedReader(resp.ReadCloser, maxMessageSize),
-		isEmpty: isEmpty,
+	return &reader[T]{
+		rc:  resp.ReadCloser,
```
liujsj, line 55: Prior we were calling the messages to a max size. What was the reason to remove it?

controller/controller.go
yushan8: oops removed

controller/getchangedtargets.go
yushan8: Right, updated.

core/storage/reader.go
yushan8: we can bump the storage version, so new instances of the service will write/read from the json entity, not the proto object.

core/storage/reader.go
yushan8: https://pkg.go.dev/github.com/gogo/protobuf/io#NewDelimitedReader protobuf reader requires a MaxSize input. `json.NewDecoder` doesn't need a max-size parameter.
Since we are already splitting the entities by bytes when writing to the storage, every `Read()` call will read at most `max_message_bytes` from the storage.

PR #224, opened 2026-07-20
feat(mapper): add entity-proto mappers

yushan8:
## Summary

Adds entity-proto mapping functions for the controller and orchestrator layers:

- `internal/mapper/` — proto↔entity converters (`GetTargetGraphResponseToProto`, `ChangedTargetsResponseToProto`, `ProtoToGetTargetGraphRequest`, `ProtoToBuildDescription`, `ToProtoError`)
- `mapper/` — top-level `ResultToTargetGraph` for converting `targethasher.Result` into ID-mapped entity types (importable by external consumers)

## Stack
1. #226 (streaming utilities)
2. **This PR** (entity-proto mappers)
3. #222 (storage JSON + entity migration)

## Test plan
- [x] `make build && make test && make gazelle`

### DIFF
--- a/mapper/mapper.go
+++ b/mapper/mapper.go
@@ -1,0 +1,97 @@
+package mapper
+
+import (
+	"context"
+	"encoding/hex"
+
+	buildpb "github.com/bazelbuild/buildtools/build_proto"
+	"github.com/uber/tango/core/targethasher"
+	"github.com/uber/tango/entity"
+	"github.com/uber/tango/internal/mapper/idmapper"
+)
+
+const cancelCheckInterval = 4096
+
+// ResultToTargetGraph converts a targethasher.Result into ID-mapped entity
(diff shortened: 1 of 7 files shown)


mapper/mapper.go
```diff
@@ -0,0 +1,97 @@
+package mapper
```
xytan0056, line 1: for https://github.com/uber/tango/pull/222#discussion_r3618704606
I meant the naming, we have 2 mapper pakage? internal/mapper and mapper

yushan8: Internally we're importing `ResultToTargetGraph` so we need the API to be visible. There could be custom orchestrator implementations that's running bazel query -> targethasher -> `ResultToTargetGraph` -> upload to storage.

xytan0056 (approved):

PR #225, opened 2026-07-21
[chore] remove cover.out after CI run

sywhang:
Local run to runs keep leaving the cover.out file, iterating becomes a little annoying.


sbalabanov (approved):
explanation would be nice to have

PR #226, opened 2026-07-21
feat(streaming): add streaming split utilities for gRPC message sizing

yushan8:
## Summary

Adds `internal/streaming` package for splitting gRPC stream messages by wire size to stay under the per-message size limit. Also adds `Size()` to `entity.OptimizedTarget` for proto wire size estimation without importing protobuf.

Target graphs are split into bounded chunks before writing to storage. When the controller reads back, each `Read()` returns one pre-split chunk that fits in a single `stream.Send()`. Without splitting, large monorepos would exceed the gRPC message size limit.

## Stack
1. **This PR** (streaming utilities)
2. #224 (entity-proto mappers, retargeted)
3. #222 (storage JSON + entity migration)

## Test plan
- [x] `make build && make test && make gazelle`


entity/optimized_target.go
```diff
@@ -14,6 +14,55 @@ type OptimizedTarget struct {
...
+		n += 1 + 1
+	}
+	for k, v := range t.Attributes {
+		entrySize := 1 + varintSize(uint64(k)) + 1 + varintSize(uint64(v))
+		n += 1 + varintSize(uint64(entrySize)) + entrySize
+	}
+	return n
+}
+
+func varintSize(x uint64) int {
```
xytan0056: duplicate in steaming.go too, can be reused

xytan0056 (approved):

PR #228, opened 2026-07-22
fix(cachekey): frame exclusion regexes

sywhang:
Length-frame sorted exclusion regexes before hashing so structurally distinct lists cannot produce the same cache key while preserving set-order independence.

Add a Rapid property test that constructs concatenation-boundary aliases, permutes both lists, and verifies distinct keys without mutating inputs.


core/cachekey/cachekey.go
```diff
@@ -75,7 +78,13 @@ func hashExcludeFilesRegex(excludeFilesRegex []string) string {
 	sort.Strings(sorted)
 	h := md5.New()
 	for _, r := range sorted {
-		h.Write([]byte(r))
+		writeFramedString(h, r)
 	}
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
+
+func writeFramedString(h hash.Hash, value string) {
```
xytan0056, line 86: smartt

xytan0056 (approved):

PR #229, opened 2026-07-22
fix(storage): preserve opaque disk keys

sywhang:
Encode disk keys under the storage root so traversal components cannot escape, normalized paths cannot alias, and parent and child keys can coexist.

Add Rapid model-based operation sequences against memory and disk storage, plus generated traversal keys guarded by an outside-root canary.


go.sum
```diff
@@ -310,6 +310,8 @@ gopkg.in/yaml.v2 v2.4.0 h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
 gopkg.in/yaml.v2 v2.4.0/go.mod h1:RDklbk79AGWmwhnvt/jBztapEOGDOx6ZbXqjP6csGnQ=
 gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
 gopkg.in/yaml.v3 v3.0.1/go.mod h1:K4uyk7z7BCEPqu6E+C64Yfv1cQ7kz7rIZviUmN+EgEM=
+pgregory.net/rapid v1.3.0 h1:vBvO0VSqti75J1jjYqpgPNBLKMd1+gxa9fYo7vk/Exc=
```
sywhang, line 313: these lines + go.sum and MODULE.bazel changes are shared between this PR and #228. Whichever one merges first will rebase.

xytan0056 (approved):

xytan0056:
actaully this makes debugging harder, we won't be able to browser the dir and find the graph cache to debug. The security fix is valid. but we can jsut strip the `../` rather than encoding the entire thing.

PR #232, opened 2026-07-22
feat(repomanager): emit lease metrics for health and slot-wait latency

xytan0056:
## What

Instruments `RepoManager.Lease` with the `observability/metrics` emitter, using the same Scope-in pattern as the orchestrator and graphrunner.

- `lease.start` counter + result-tagged (`success`/`failure`/`cancelled`) `lease.finish` histogram — pool health.
- Per-step latency histograms: `ensure_origin_duration`, `wait_slot_duration` (the worker-slot wait — the saturation signal), `create_worker_duration`.

Repo-tagged; buckets are exponential 1ms..~1.3h. The example server leaves the scope no-op (as the orchestrator/graphrunner are wired there today); production wiring supplies a reporter-backed scope.

Also folds nil-scope handling into `metrics.New` (nil → no-op, no error) and adds `Emitter.SubScope`, removing the repeated `if scope == nil` / `if err != nil { Nop() }` boilerplate across all call sites (repomanager, controller, graphrunner, orchestrator).

## Test

`make build` / `make test` / `make gazelle` pass.


core/repomanager/repo_manager.go
```diff
@@ -98,11 +105,20 @@ func NewRepoManager(appCtx context.Context, p Params) (RepoManager, error) {
 	if p.PoolSize <= 0 {
 		return nil, fmt.Errorf("pool size must be > 0, got %d", p.PoolSize)
 	}
+	scope := p.Scope
+	if scope == nil {
+		scope = tally.NoopScope
+	}
+	emitter, err := metrics.New(scope.SubScope("repo_manager"))
+	if err != nil {
+		emitter = metrics.Nop()
```
sywhang: can we return this error instead of silently dropping metrics?

core/repomanager/repo_manager.go
```diff
@@ -140,19 +156,29 @@ func (r *repoManager) poolFor(repo string) *workerPool {
...
+	op := metrics.Begin(e, _opLease, _stepDurationBuckets)
+	defer func() { op.Complete(retErr) }()
+
 	pool := r.poolFor(repo)
 
-	if err := pool.ensureOrigin(ctx, r.git, desc.Remote); err != nil {
+	originStart := time.Now()
+	err := pool.ensureOrigin(ctx, r.git, desc.Remote)
+	recordStep(e, _stepEnsureOrigin, originStart)
+	if err != nil {
```
yushan8, line 162: nit: wrap error so we can trace where it came from.
```suggestion
		return nil, fmt.Errorf("ensure origin: %w", err)
```

yushan8 (approved):

core/repomanager/repo_manager.go
yushan8: It seems like we're always checking outside of the function `if scope == nil` https://github.com/uber/tango/blob/main/graphrunner/native.go#L49. Should this check be in `metrics.New` to set `scope = tally.NoopScope`?
We probably just want to make sure the initialization is consistent. Either we return an error if it's not provided or fallback to `scope = tally.NoopScope` if scope == nil.

core/repomanager/repo_manager.go
```diff
@@ -140,19 +156,29 @@ func (r *repoManager) poolFor(repo string) *workerPool {
...
 	}
 
 	// Acquire a worker slot (blocks if all slots are leased)
+	waitStart := time.Now()
 	var slot *workerSlot
 	select {
 	case slot = <-pool.avail:
+		recordStep(e, _stepWaitSlot, waitStart)
 	case <-ctx.Done():
+		recordStep(e, _stepWaitSlot, waitStart)
```
liujsj: To remove the duplicated lines, would it be better to just track an error when ctx.Done() and return outside of the select block?

core/repomanager/repo_manager.go
xytan0056: rethought about this, metrics.New shouldn't return error for nil scope, should just fall back to noop

PR #235, opened 2026-07-23
feat(git): classify git command timeouts as infra errors

yushan8:
## Summary
Intent:
- Git commands (checkout, fetch, clone, diff, apply patch, commit,
  submodule update) already run under a fixed timeout, but a timeout
  failure was indistinguishable from any other git error or context.Cancelled error.

Changes:
- Added a sentinel `git.ErrTimeout` and a `wrapError` helper in
  `core/git` that wraps a failing git operation's error with its name,
  and additionally wraps ErrTimeout when the failure was caused by the
  command's own timeout (as opposed to a parent context cancellation).
- Added `orchestrator.classifyGitError`, used at the checkout,
  apply-requests, and treehash-compute call sites in
  `native_orchestrator.go`, which classifies the error as
  `tangoerrors.NewInfra` when it wraps `git.ErrTimeout`, per the
  classification conventions in `docs/errors/errors.md`.

## Test Plan
unit test

### DIFF
--- a/orchestrator/errors.go
+++ b/orchestrator/errors.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 
 	tangoerrors "github.com/uber/tango/core/errors"
-	"github.com/uber/tango/core/git"
 	"github.com/uber/tango/core/repomanager"
 )
 
@@ -31,11 +30,9 @@ func classifyLeaseError(err error) error {
 	return tangoerrors.NewInfra(wrappedErr)
 }
 
-// classifyGitError classifies err as an infra error when it was caused by a
-// git command timing out.
+// classifyGitError classifies a git-command failure as an infra error: git
+// failures (including timeouts) are always infrastructure issues, not
+// caused by user input.
 func classifyGitError(err error) error {
-	if errors.Is(err, git.ErrTimeout) {
-		return tangoerrors.NewInfra(err)
-	}
-	return err
+	return tangoerrors.NewInfra(err)
 }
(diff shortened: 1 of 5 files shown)


orchestrator/native_orchestrator.go
```diff
@@ -138,7 +138,7 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, req entity.GetT
 	err = ws.Checkout(ctx, build.Remote, build.BaseSha)
 	recordStep(e, "checkout_duration", checkoutStart)
 	if err != nil {
-		return nil, fmt.Errorf("checkout %s@%s: %w", build.Remote, build.BaseSha, err)
+		return nil, classifyGitError(err)
```
xytan0056: is it inevitable that we'd lose the `checkout %s@%s` prefix? Wondering if this info is still useful in the error.

orchestrator/errors.go
```diff
@@ -29,3 +30,12 @@ func classifyLeaseError(err error) error {
 	}
 	return tangoerrors.NewInfra(wrappedErr)
 }
+
+// classifyGitError classifies err as an infra error when it was caused by a
+// git command timing out.
+func classifyGitError(err error) error {
+	if errors.Is(err, git.ErrTimeout) {
```
xytan0056: if the error is `"exit status 128"`, then it wouldn't be classified as NewInfra. It should be infra? The unwrapped error if not timeout, would be treated as infra by `tangoerrors.GetErrorCode`, becuase it just happens to be unclassfied.

yushan8: Yea it should be `NewInfra`. I can classify them in this PR.

xytan0056 (approved):

PR #236, opened 2026-07-24
fix(bazel): classify bazelisk download network failures as retryable

yushan8:
## Summary
Intent:
- A bazel client creation failure previously returned a plain error that
  fell through to the ErrorInfra default in tangoerrors.GetErrorCode --
  correct by accident, not by design. A transient network failure while
  downloading bazelisk is retryable and shouldn't be lumped in with
  permanent infra failures (e.g. a bad BazelCommand path).

Changes:
- Add bazel.ErrDownloadBazeliskNetwork, wrapped when downloading bazelisk
  fails with a net.Error.
- Add orchestrator.classifyBazelClientError, classifying
  ErrDownloadBazeliskNetwork as tangoerrors.ErrorInfraRetryable and
  everything else as tangoerrors.ErrorInfra; wired into nativeOrchestrator.

## Test Plan
unit test

## Revert Plan
Revert this PR; classification falls back to the existing ErrorInfra default.

## Issues

### DIFF
--- a/core/bazel/bazel.go
+++ b/core/bazel/bazel.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -34,6 +35,11 @@ const (
 	// default query timeout if not provided in config
 	_queryTimeout = 15 * time.Minute
 )
+
+// ErrDownloadBazeliskNetwork indicates the bazelisk binary could not be
+// downloaded due to a transient network failure, as opposed to a permanent
+// failure (e.g. a non-200 response). Retrying is expected to succeed.
+var ErrDownloadBazeliskNetwork = errors.New("download bazelisk network failure")
 
 type QueryRequest struct {
 	Query          string
@@ -137,6 +143,10 @@ func ensureBazelisk(ctx context.Context) (_ string, retErr error) {
 	}
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
+		var netErr net.Error
+		if errors.As(err, &netErr) {
+			return "", fmt.Errorf("download bazelisk: %w: %w", ErrDownloadBazeliskNetwork, err)
+		}
 		return "", fmt.Errorf("download bazelisk: %w", err)
 	}
 	defer func() { _ = resp.Body.Close() }()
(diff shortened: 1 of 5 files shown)


graphrunner/native.go
```diff
@@ -78,7 +78,7 @@ func (g *nativeGraphRunner) Compute(ctx context.Context, ws workspace.Workspace)
 	})
 	g.emitter.DurationHistogram(_opCompute, "bazel_query_duration", _phaseDurationBuckets).RecordDuration(time.Since(bazelStart))
 	if err != nil {
-		return targethasher.EmptyResult(), err
+		return targethasher.EmptyResult(), classifyBazelQueryError(err)
```
xytan0056: looks simple enough to inline ?
```
return targethasher.Empty, tangoerrors.NewInfra(error.Wrap(err, "baze query fail..."))
```

yushan8: Updated. We have different errors based on the exit code so this needs to be wrapped in a function to classify them.

core/bazel/query.go
```diff
@@ -95,6 +121,9 @@ func (b *BazelClient) executeQueryInternal(ctx context.Context, query string, st
 	waitErr := cmd.Wait()
 	streamErr := g.Wait()
 	if waitErr != nil {
+		if isGraphFailure(waitErr) {
+			return queryResults, b.wrapQueryFailure("bazel query failed", fmt.Errorf("%w: %w", ErrQueryGraphFailure, waitErr), &stderrBuf)
```
xytan0056: @justinwon777 
CLI imports this pacakge , does this error surface to bk jobs (end up a specific error code) and get classified accordingly? or should we let uber's bk classifier take over everything?

core/bazel/bazel.go
```diff
@@ -137,6 +143,10 @@ func ensureBazelisk(ctx context.Context) (_ string, retErr error) {
 	}
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
+		var netErr net.Error
+		if errors.As(err, &netErr) {
+			return "", fmt.Errorf("download bazelisk: %w: %w", ErrDownloadBazeliskNetwork, err)
```
justinwon777: this will say "download bazelisk: download bazelisk network failure". Can you make it not repeat "download bazelisk"?

justinwon777: `ErrDownloadBazeliskNetwork` could be more generic like `ErrNetwork` saying "network failure" since it's when error is net.Error, not necessarily bazelisk related.

yushan8: Sg, updated.

xytan0056 (approved):

PR #239, opened 2026-07-24
fix(cachekey): include ordered request commits

sywhang:
This is a similar patch to #228, against internal/url. 

Include each change request URL and commit in caller order with length framing so distinct build descriptions cannot share a treehash cache entry.

Add Rapid properties asserting that changing a commit or reordering change requests changes the cache path.


internal/url/url.go
```diff
@@ -30,20 +32,22 @@ func ToShortRemote(remote string) string {
...
-	for _, url := range urls {
-		h.Write([]byte(url))
+	for _, request := range requests {
+		writeFramedString(h, request.URL)
+		writeFramedString(h, request.Commit)
 	}
 	return fmt.Sprintf("%x", h.Sum(nil))
 }
+
+func writeFramedString(h hash.Hash, value string) {
```
xytan0056, line 49: worth dedupe from the one in  cachekey https://github.com/uber/tango/blob/main/core/cachekey/cachekey.go#L86

xytan0056 (approved):

PR #241, opened 2026-07-24
feat(controller): emit result-tagged cache-lookup counters

xytan0056:
Replace the hit-only `cache_hit` counters with `result`-tagged cache-lookup counters at every controller cache-lookup site, so a per-layer hit rate (`hit / (hit + miss)`) is derivable for each cache.

Three counters, each emitted under its parent RPC op with a `result=hit|miss` tag:

| metric | emitted from |
|---|---|
| `treehash_cache_lookup` | `get_target_graph`, `get_changed_targets` |
| `graph_cache_lookup` | `get_target_graph` |
| `compared_targets_cache_lookup` | `get_changed_targets` |

## Classification

The result is classified by the lookup error's **origin**: a not-found is a genuine miss; an infra failure is *not* a cache miss (it's already tracked by the failure metric) and is left out so it can't skew the hit rate.

| site | not-found | infra error |
|---|---|---|
| treehash | miss | nothing (request fails) |
| graph | miss | nothing (request fails) |
| compared-targets | miss | nothing (recomputes) |

**Breaking:** the bare `controller.get_target_graph.cache_hit` and `controller.get_changed_targets.cache_hit` counters are removed.

Stacked on `refactor/share-metrics-buckets` (#240) — review/merge that first.

### DIFF
--- a/controller/gettargetgraph.go
+++ b/controller/gettargetgraph.go
@@ -103,6 +103,7 @@ func (c *controller) getGraph(ctx context.Context, e *metrics.Emitter, req entit
 		// Look up the the git treehash based on cache path
 		treehashCachePath := cachekey.GetTreehashCachePath(req.Build)
 		treehashResponse, err := c.storage.Get(ctx, storage.DownloadRequest{Key: treehashCachePath})
+		recordCacheLookup(e, opGetTargetGraph, _metricTreehashCacheLookup, err)
 		if err != nil {
 			if storage.IsNotFound(err) {
 				// Cache miss - blob doesn't exist, need to compute and store target graph
@@ -122,10 +123,11 @@ func (c *controller) getGraph(ctx context.Context, e *metrics.Emitter, req entit
 			// Download the target graph based on treehash.
 			storageStart := time.Now()
 			graphReader, err := storage.NewGraphReader(ctx, c.storage, treehashPath)
+			if ctx.Err() != nil {
+				err = ctx.Err()
+			}
+			recordCacheLookup(e, opGetTargetGraph, _metricGraphCacheLookup, err)
 			if err != nil {
-				if ctx.Err() != nil {
-					err = ctx.Err()
-				}
 				if !storage.IsNotFound(err) {
 					return nil, fmt.Errorf("graph reader: %w", err)
 				}
@@ -135,7 +137,6 @@ func (c *controller) getGraph(ctx context.Context, e *metrics.Emitter, req entit
 					zap.Duration("storage_duration", time.Since(storageStart)),
 					zap.Duration("total_duration", time.Since(start)),
 				)
-				e.Counter(opGetTargetGraph, "cache_hit").Inc(1)
 				e.DurationHistogram(opGetTargetGraph, "download_graph", metrics.SlowDurationBuckets).RecordDuration(time.Since(storageStart))
 				return graphReader, nil
 			}
--- a/controller/metrics.go
+++ b/controller/metrics.go
@@ -14,6 +14,11 @@
 
 package controller
 
+import (
+	"github.com/uber/tango/core/storage"
+	"github.com/uber/tango/observability/metrics"
+)
+
 // Operation names, snake_cased after the RPC interface methods they measure.
 const (
 	opGetTargetGraph        = "get_target_graph"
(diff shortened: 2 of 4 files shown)


controller/gettargetgraph.go
```diff
@@ -122,10 +123,11 @@ func (c *controller) getGraph(ctx context.Context, e *metrics.Emitter, req entit
 			// Download the target graph based on treehash.
 			storageStart := time.Now()
 			graphReader, err := storage.NewGraphReader(ctx, c.storage, treehashPath)
+			if err != nil && ctx.Err() != nil {
```
justinwon777: i dont think you need to check err != nil. No matter what if ctx.Err() I= nil we want to set err = ctx.Err()

controller/metrics.go
```diff
@@ -14,10 +14,31 @@
...
+const (
+	metricTreehashCacheLookup        = "treehash_cache_lookup"
+	metricGraphCacheLookup           = "graph_cache_lookup"
+	metricComparedTargetsCacheLookup = "compared_targets_cache_lookup"
+)
+
+// recordCacheLookup emits a result-tagged counter for a cache lookup under the
+// given parent op: hit when the payload was usable, miss otherwise (not-found
+// or a storage failure both mean the payload must be recomputed).
+func recordCacheLookup(e *metrics.Emitter, parentOp, name string, hit bool) {
```
justinwon777: can this take in err instead of hit so you can check if err == nil here instead of doing it at all the call sites?

controller/metrics.go
```diff
@@ -14,10 +14,31 @@
...
 	opGetChangedTargetGraph = "get_changed_target_graph"
 	opCompareTargetGraphs   = "compare_target_graphs"
 )
+
+// Cache-lookup metric names. Each is emitted under its parent RPC op with a
+// result=hit|miss tag, so a hit rate is derivable per cache layer.
+const (
+	metricTreehashCacheLookup        = "treehash_cache_lookup"
+	metricGraphCacheLookup           = "graph_cache_lookup"
+	metricComparedTargetsCacheLookup = "compared_targets_cache_lookup"
```
yushan8:
```suggestion
	_metricTreehashCacheLookup        = "treehash_cache_lookup"
	_metricGraphCacheLookup           = "graph_cache_lookup"
	_metricComparedTargetsCacheLookup = "compared_targets_cache_lookup"
```

controller/gettargetgraph.go
```diff
@@ -122,10 +123,11 @@ func (c *controller) getGraph(ctx context.Context, e *metrics.Emitter, req entit
...
+				err = ctx.Err()
+			}
+			recordCacheLookup(e, opGetTargetGraph, metricGraphCacheLookup, err == nil)
```
yushan8: nit: Maybe we should do a check for `if err == nil { cacheHit = true }`
And then pass that into `recordCacheLookup(e, opGetTargetGraph, metricGraphCacheLookup, cacheHit)` for readability purposes. It's hard to tell why err == nil is passed into the function param of `hit`.
Also I think we should only emit this metric if storage.IsNotFound(err) in line 134 and line 136 separately. If it's neither a `NotFoundError` or err == nil, should it even emit a cache hit/miss metric?

controller/metrics.go
yushan8: I think we only care if it's a notfounderror? Any other error I don't think we should emit this metric because it's an infra failure then?

yushan8 (approved):

controller/metrics.go
xytan0056: > check if err == nil here
that's what I did initially--passing the error in, but there are cases when there's no error to pass like L196
```
recordCacheLookup(e, opGetChangedTargets, metricComparedTargetsCacheLookup, true)
```
I can pass in the error itself here

controller/gettargetgraph.go
xytan0056: make sense. originally I was thinking all errors should be cache miss, but now i'm thinking cache hit should be a effectiveness meassurement not an outcome. The cachehit metrics should be honest here.
The only downside is `hits + misses ≠ total attempts`

justinwon777 (approved):

PR #246, opened 2026-07-28
fix(bazel): identify bazel query context timeouts wrapping the error

yushan8:
## Summary
Wrap the error if bazel query timed out so we can identify which command was attributed to `context.DeadlineExceeded`.

Follow-up to #236, which classified bazelisk download network failures as
retryable (a different case: that one is a genuine transient failure
before any query even starts).

## Test Plan
- Added TestExecuteQueryInternal_ContextTimeout assertion that the error
  now wraps bazel.ErrQueryTimeout.
- Added TestExecuteQueryInternal_WaitFailureWithoutTimeout to confirm a
  plain wait failure (no context deadline) is not misclassified as a
  timeout.
- Added TestExecuteQueryInternal_StreamTimeoutWithoutWaitError to confirm
  a timeout surfacing only via the stream-reading goroutines is still
  wrapped with ErrQueryTimeout.

## Revert Plan
Revert this PR; the sentinel and its detection are removed, with no
change to runtime classification behavior.

### DIFF
--- a/core/bazel/query.go
+++ b/core/bazel/query.go
@@ -106,14 +106,15 @@ func (b *BazelClient) executeQueryInternal(ctx context.Context, query string, st
 
 // wrapQueryFailure logs the failure and returns a wrapped error. ctx is the
 // query's own timeout context (not a parent cancellation): if it has hit its
-// deadline, "query timeout exceeded" is noted in the message so this failure
-// can be told apart from the caller disconnecting, without a dedicated
-// sentinel error. When stderr was captured (streamLogs off), its contents are
-// appended so the failure is self-contained. When streamLogs is on the
-// operator has already seen stderr live, so it's omitted.
+// deadline, cause is additionally wrapped with ctx.Err() (context.DeadlineExceeded)
+// so callers can identify via errors.Is that this failure was the query's own
+// timeout elapsing, as opposed to the caller disconnecting, without a
+// dedicated sentinel error. When stderr was captured (streamLogs off), its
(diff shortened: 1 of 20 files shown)


core/bazel/query.go
```diff
@@ -95,20 +95,27 @@ func (b *BazelClient) executeQueryInternal(ctx context.Context, query string, st
...
+// query's own timeout context (not a parent cancellation): if it has hit its
+// deadline, cause is additionally wrapped with ErrQueryTimeout so callers can
+// identify via errors.Is that this failure was the query's own timeout
+// elapsing, as opposed to the caller disconnecting. When stderr was captured
+// (streamLogs off), its contents are appended so the failure is
 // self-contained. When streamLogs is on the operator has already seen stderr
 // live, so it's omitted.
-func (b *BazelClient) wrapQueryFailure(msg string, cause error, stderrBuf *bytes.Buffer) error {
+func (b *BazelClient) wrapQueryFailure(ctx context.Context, msg string, cause error, stderrBuf *bytes.Buffer) error {
+	if ctx.Err() == context.DeadlineExceeded {
```
justinwon777: we should do this check, but a sentinel seems overkill since we're not handling it anywhere. A plain `fmt.Errorf` achieves the same result. Unless we plan to do something with it right now, I'd prefer just making it a plain error until we specifically want to handle this.

yushan8: That's fair. Updated.

justinwon777 (approved):

PR #248, opened 2026-07-30
fix(bazel): make stream handling synchronous to eliminate timeout race (audit #2)

sbalabanov:
## Summary

Data race in Bazel timeout handling: `streamOutput` and `streamAndParseTargets` launched nested goroutines and returned immediately on context cancellation, allowing `executeQueryInternal` to read buffers while detached stream work could still be writing.

- Remove the nested goroutines and block synchronously on `io.Copy` / `getQueryResult` inside the existing errgroup workers.
- After each blocking operation returns, prefer `context.Cause(ctx)` so timeout and cancellation semantics remain intact.
- No context reader wrapper is needed for these command pipes: `exec.CommandContext` cancellation terminates the subprocess and `Wait` closes the stdout/stderr pipes, unblocking reads. `getQueryResult` has the same pipe behavior and also polls the context while parsing targets.

## Test plan

- [x] `go test -race -count=5 ./core/bazel`
- [x] `go test ./core/bazel`
- [x] `go build ./...`

### DIFF
--- a/core/bazel/stream.go
+++ b/core/bazel/stream.go
@@ -24,38 +24,19 @@ import (
 )
 
 func streamOutput(ctx context.Context, src io.Reader, dst io.Writer) error {
-	done := make(chan error, 1)
-	go func() {
-		_, err := io.Copy(dst, src)
-		done <- err
-	}()
-
-	select {
-	case <-ctx.Done():
-		return context.Cause(ctx)
-	case err := <-done:
-		return err
+	_, err := io.Copy(dst, src)
+	if cause := context.Cause(ctx); cause != nil {
+		return cause
 	}
+	return err
 }
 
 func streamAndParseTargets(ctx context.Context, src io.Reader, dst io.Writer) (*buildpb.QueryResult, error) {
-	type result struct {
-		queryResult *buildpb.QueryResult
-		err         error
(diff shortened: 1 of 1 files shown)


sbalabanov (commented):
Rework the approach. There is no need for a separate goroutine to do copy or read query results - if they properly respond to cancellations then the implementation can just block on them in a main thread.
Inspect if io.Copy will properly respond to cancellation, if not then use ctxReader wrapper.
Inspect getQueryResult() as well and if it does not, suggest the fix but do not do it.
Both functions need not to have goroutines anymore.

sbalabanov:
Addressed the review in 0ae6f38cf7cfa684d38c1e2a8d01d96d0721b249 by removing both nested stream goroutines. The existing errgroup workers now block directly on io.Copy and getQueryResult, then return the context cause when cancellation occurred.

I also checked cancellation behavior: these readers are stdout/stderr pipes from exec.CommandContext, so cancellation terminates the subprocess and Wait closes the pipes, which unblocks both io.Copy and protodelim parsing. getQueryResult additionally polls the context while processing targets, so no separate ctxReader or follow-up change is needed for this call path.

[addressed by agent]

xytan0056 (approved):

PR #249, opened 2026-07-30
fix(orchestrator): make treehash-mapping cache write best-effort (audit #8)

sbalabanov:
## Summary

The treehash-mapping `Put` is a pure cache optimization (maps build description to treehash to skip workspace materialization on future requests), but its failure was incorrectly failing the entire request after a successful graph computation.

- Make treehash-mapping `Put` best-effort: on error, log a structured warning and continue serving the computed result
- Add inline comment documenting the two-tier cache write policy (identity-bearing writes are load-bearing; pure-optimization writes are best-effort)
- Add table-driven tests pinning both behaviors: treehash Put failure still succeeds, graph blob write failure still fails

### DIFF
--- a/orchestrator/BUILD.bazel
+++ b/orchestrator/BUILD.bazel
@@ -40,6 +40,7 @@ go_test(
     deps = [
         "//config",
         "//core/bazel",
+        "//core/cachekey",
         "//core/errors",
         "//core/git",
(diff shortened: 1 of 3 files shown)


xytan0056 (approved):

sbalabanov:
we prefer to fail fast instead

PR #250, opened 2026-07-30
fix(controller): return classified TangoError details to RPC clients (audit #4)

sbalabanov:
## Summary

RPC handlers returned plain wrapped errors, so clients never received the documented `TangoError` detail on the wire despite `internal/mapper.ToProtoError` being implemented. This violated the proto contract in `docs/errors/errors.md`.

- Added `toWireError` helper in `controller/errors.go` that delegates to `mapper.ToProtoError`
- Every handler's existing named-return defer now calls `retErr = toWireError(retErr)` before returning, ensuring uniform coverage
- Added table-driven and end-to-end tests verifying user-classified and infra-classified errors surface as YARPC errors carrying the correct `TangoError` detail (`ERROR_USER`, `ERROR_INFRA`, etc.)
- Updated `docs/errors/errors.md` to reflect the actual defer-based mechanism

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./controller/` passes (including new wire error tests)
- [x] `go test ./...` passes (integration tests excluded -- require env var)
- [x] `make gazelle` picks up new test file and YARPC deps

### DIFF
--- a/controller/BUILD.bazel
+++ b/controller/BUILD.bazel
@@ -45,6 +45,7 @@ go_test(
         "metrics_test.go",
         "output_filter_test.go",
         "testhelper_test.go",
+        "wireerror_test.go",
     ],
     embed = [":controller"],
     deps = [
@@ -61,6 +62,8 @@ go_test(
         "@com_github_stretchr_testify//require",
         "@com_github_uber_go_tally//:tally",
         "@org_uber_go_mock//gomock",
+        "@org_uber_go_yarpc//encoding/protobuf",
+        "@org_uber_go_yarpc//yarpcerrors",
         "@org_uber_go_zap//:zap",
         "@org_uber_go_zap//zaptest",
     ],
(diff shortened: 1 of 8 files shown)


controller/getchangedtargetgraph.go
```diff
@@ -24,6 +24,11 @@ import (
 // stub: it records the start/finish lifecycle and returns no data.
 func (c *controller) GetChangedTargetGraph(request *pb.GetChangedTargetGraphRequest, stream pb.TangoServiceGetChangedTargetGraphYARPCServer) (retErr error) {
 	op := metrics.Begin(c.emitter, opGetChangedTargetGraph, metrics.SlowDurationBuckets)
-	defer func() { op.Complete(retErr) }()
+	defer func() {
+		op.Complete(retErr)
+		if retErr != nil {
```
xytan0056: > Nil errors pass through unchanged
no need to check nil

xytan0056 (approved):

controller/getchangedtargetgraph.go
sbalabanov: I noticed that when I did code review myself, but decided it is more readable this way, as nothing in the name of of the wireError says that it bypasses nils.

PR #251, opened 2026-07-30
fix(errors): classify shutdown as cancelled and user-origin failures as user (audit #10)

sbalabanov:
## Summary

App-shutdown cancellation was misclassified as `infra` because the cancel cause was a plain error that did not wrap `context.Canceled`. Additionally, unknown-repo and graph-runner/Bazel computation failures were returned without classification, defaulting to `infra` even when caused by client input.

- **linkRequestCtx**: cancel cause now wraps `context.Canceled` so `GetErrorCode` returns `ErrorCancelled`
- **Unknown repository**: classified as `ErrorUser` since it is caused by client input
- **Runner/Bazel computation failures**: new `classifyRunnerError` defaults to `ErrorUser` (broken BUILD files) unless a recognized infra sentinel matches (`bazel.ErrNetwork`, `git.ErrTimeout`, `git.ErrFatal`, `context.DeadlineExceeded`)
- Tests extended in `orchestrator/errors_test.go` and updated in `controller/controller_test.go`

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -16,7 +16,7 @@ package controller
 
 import (
 	"context"
-	"errors"
+	"fmt"
 
 	"github.com/uber-go/tally"
 	"github.com/uber/tango/config"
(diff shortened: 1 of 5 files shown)


xytan0056 (approved):

sbalabanov:
Incorrect: infra should be the default and not user.
Classification may be needed to be done earlier in the calling stack.

PR #252, opened 2026-07-30
fix(itg): encode remotes in filesystem-safe cache keys (audit #13)

sbalabanov:
## Summary

Encode repository remotes into a slash-free component before constructing ITG cache keys. This lets `toStorageKey` and `FloorKey` use the same `path.Join` layout without URL or SSH remotes accidentally creating filesystem path segments.

## Intent

Make ITG cache keys portable across storage backends, including filesystem-based implementations, while ensuring writes, reads, and prefix listings derive identical keys for URL, path, and SSH remotes.

## Changes

- Percent-encode reserved path characters while preserving readable remote names.
- Build cache entry keys and listing prefixes with `path.Join`.
- Add table-driven round-trip coverage for URL, plain-path, and SSH remotes.
- Verify the stored remote component is readable, slash-free, and decodes to the original remote.
- Treat existing entries under the old key format as cold cache data; they will be regenerated rather than migrated.

## Test Plan

- [x] `go test ./core/itg/cache/...`
- [x] `./tools/bazel test //core/itg/cache:cache_test --test_output=errors`
- [x] `make gazelle`
- [x] `aifx verify`
- [x] `go test ./...` passes outside integration tests; integration requires `TANGO_REPO_REMOTE`

## Revert Plan

Revert the commits from this PR to restore the previous ITG cache-key format.

## Jira Issues

None.

### DIFF
--- a/core/itg/cache/cache.go
+++ b/core/itg/cache/cache.go
@@ -17,8 +17,10 @@ package cache
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/gob"
 	"fmt"
+	"path"
 	"slices"
 	"strconv"
 	"strings"
@@ -28,7 +30,7 @@ import (
 	"github.com/uber/tango/core/storage"
 )
 
-const keyPrefix = "itg/"
+const keyPrefix = "itg"
 
 // Cache is an interface for caching optimized graphs.
 type Cache interface {
@@ -63,15 +65,16 @@ var CompareKeyFunc = func(a Key, b Key) int {
 // EmptyKey means no cache found.
 var EmptyKey = Key{}
 
-// toStorageKey converts a cache key to its storage key: itg/{remote}/{date}/{committime}_{sha}
-//
-// Keys are built with string concatenation, not filepath.Join, because storage
-// keys are opaque strings and filepath.Join would collapse double slashes in
-// URL-style remotes (e.g. "https://github.com/x" -> "https:/github.com/x")
-// and use platform-dependent separators.
+// toStorageKey converts a cache key to its storage key:
+// itg/{encoded_remote}/{date}/{committime}_{sha}.
 func (k *Key) toStorageKey() string {
 	date := time.Unix(k.BaseCommitTimeSecond, 0).UTC().Format("2006-01-02")
-	return keyPrefix + k.Remote + "/" + date + "/" + fmt.Sprintf("%d_%s", k.BaseCommitTimeSecond, k.BaseSha)
+	return path.Join(keyPrefix, encodeRemote(k.Remote), date, fmt.Sprintf("%d_%s", k.BaseCommitTimeSecond, k.BaseSha))
+}
+
+// encodeRemote returns a filesystem-safe, reversible remote key component.
+func encodeRemote(remote string) string {
+	return base64.RawURLEncoding.EncodeToString([]byte(remote))
 }
 
 // NewStorageCache creates a new cache backed by a storage.Storage implementation.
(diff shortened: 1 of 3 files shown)


xytan0056:
ITG local cache is designed for disk path lookup.

core/itg/cache/cache.go
```diff
@@ -65,9 +64,14 @@ var CompareKeyFunc = func(a Key, b Key) int {
...
 // toStorageKey converts a cache key to its storage key: itg/{remote}/{date}/{committime}_{sha}
+//
+// Keys are built with string concatenation, not filepath.Join, because storage
+// keys are opaque strings and filepath.Join would collapse double slashes in
+// URL-style remotes (e.g. "https://github.com/x" -> "https:/github.com/x")
+// and use platform-dependent separators.
 func (k *Key) toStorageKey() string {
 	date := time.Unix(k.BaseCommitTimeSecond, 0).UTC().Format("2006-01-02")
-	return filepath.Join(keyPrefix, k.Remote, date, fmt.Sprintf("%d_%s", k.BaseCommitTimeSecond, k.BaseSha))
+	return keyPrefix + k.Remote + "/" + date + "/" + fmt.Sprintf("%d_%s", k.BaseCommitTimeSecond, k.BaseSha)
```
sbalabanov: encode "remote" instead so it won't contain slashes. The goal is to make the key compatible to the common file-system based storage.
after that you can use path.join() which always uses forward slashes.

sbalabanov: Encoded the remote with base64.RawURLEncoding and switched both entry keys and FloorKey prefixes to path.Join in f53448d715d11029f426abce407392127cb3f212. This keeps the remote in one slash-free, filesystem-safe component while preserving deterministic lookup behavior; the round-trip tests now verify the component decodes to the original URL, path, or SSH remote.

[addressed by agent]

core/itg/cache/cache.go
```diff
@@ -64,10 +65,16 @@ var CompareKeyFunc = func(a Key, b Key) int {
...
+// itg/{encoded_remote}/{date}/{committime}_{sha}.
 func (k *Key) toStorageKey() string {
 	date := time.Unix(k.BaseCommitTimeSecond, 0).UTC().Format("2006-01-02")
-	return filepath.Join(keyPrefix, k.Remote, date, fmt.Sprintf("%d_%s", k.BaseCommitTimeSecond, k.BaseSha))
+	return path.Join(keyPrefix, encodeRemote(k.Remote), date, fmt.Sprintf("%d_%s", k.BaseCommitTimeSecond, k.BaseSha))
+}
+
+// encodeRemote returns a filesystem-safe, reversible remote key component.
+func encodeRemote(remote string) string {
+	return base64.RawURLEncoding.EncodeToString([]byte(remote))
```
sbalabanov: encode so that when name characters can be preserved, they are preserved, base64 hides the name

sbalabanov:
> ITG local cache is designed for disk path lookup.

Nothing in the ITG cache contract says that. The very general ITG cache needs to be distributed, with local caching as an optimization.

core/itg/cache/cache.go
sbalabanov: Switched remote encoding from Base64 to URL path escaping in 1e95661671b11b06980753c5ec031b8d322d9aff. This preserves recognizable remote characters while escaping each slash as %2F, so the remote remains one reversible path component. The tests now assert the exact readable encoding and round-trip decoding.

[addressed by agent]

xytan0056 (approved):

PR #253, opened 2026-07-30
fix(controller): return Unimplemented from GetChangedTargetGraph stub (audit #5)

sbalabanov:
## Summary

`GetChangedTargetGraph` was a silent stub that returned nil (empty success stream) while the proto contract promises a full streamed changed target graph. Metrics reported success, making it impossible to distinguish the stub from a real call.

- Return a YARPC `Unimplemented` error classified as `ErrorUser` so clients get an explicit failure
- Metrics lifecycle now reflects the failure: finish histogram tagged `result=user`, failure counter emitted with `error_code=user`
- Proto RPC comment annotated with "NOT YET IMPLEMENTED"
- Regenerated tangopbmock to include `TangoServiceGetChangedTargetGraphYARPCServer`
- Added table-driven test asserting the YARPC error code and that no chunks are streamed
- Updated existing metrics path-shape test to expect the new failure outcome

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./controller/...` passes (new + updated tests)
- [x] `make gazelle` BUILD files in sync

### DIFF
--- a/controller/getchangedtargetgraph.go
+++ b/controller/getchangedtargetgraph.go
@@ -15,15 +15,24 @@
 package controller
 
 import (
+	tangoerrors "github.com/uber/tango/core/errors"
 	"github.com/uber/tango/observability/metrics"
 	pb "github.com/uber/tango/tangopb"
+	"go.uber.org/yarpc/yarpcerrors"
 )
 
 // GetChangedTargetGraph is the streaming RPC that will return the subgraph
-// induced by the changed targets between two revisions. It is currently a
-// stub: it records the start/finish lifecycle and returns no data.
+// induced by the changed targets between two revisions. NOT YET IMPLEMENTED:
+// returns a YARPC Unimplemented error so callers get an explicit failure
+// instead of a silent empty stream.
 func (c *controller) GetChangedTargetGraph(request *pb.GetChangedTargetGraphRequest, stream pb.TangoServiceGetChangedTargetGraphYARPCServer) (retErr error) {
 	op := metrics.Begin(c.emitter, opGetChangedTargetGraph, metrics.SlowDurationBuckets)
-	defer func() { op.Complete(retErr) }()
-	return nil
+	retErr = tangoerrors.NewInfra(
+		yarpcerrors.Newf(yarpcerrors.CodeUnimplemented, "GetChangedTargetGraph is not yet implemented"),
+	)
+	defer func() {
+		op.Complete(retErr)
+		emitFailureMetric(c.emitter, opGetChangedTargetGraph, retErr)
+	}()
+	return retErr
 }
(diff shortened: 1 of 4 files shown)


controller/getchangedtargetgraph.go
```diff
@@ -15,15 +15,24 @@
...
-// induced by the changed targets between two revisions. It is currently a
-// stub: it records the start/finish lifecycle and returns no data.
+// induced by the changed targets between two revisions. NOT YET IMPLEMENTED:
+// returns a YARPC Unimplemented error so callers get an explicit failure
+// instead of a silent empty stream.
 func (c *controller) GetChangedTargetGraph(request *pb.GetChangedTargetGraphRequest, stream pb.TangoServiceGetChangedTargetGraphYARPCServer) (retErr error) {
 	op := metrics.Begin(c.emitter, opGetChangedTargetGraph, metrics.SlowDurationBuckets)
-	defer func() { op.Complete(retErr) }()
-	return nil
+	retErr = tangoerrors.NewUser(
```
sbalabanov: should be infra error

controller/getchangedtargetgraph_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	pb "github.com/uber/tango/tangopb"
+	tangomock "github.com/uber/tango/tangopb/tangopbmock"
+	"go.uber.org/mock/gomock"
+	"go.uber.org/yarpc/yarpcerrors"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestGetChangedTargetGraph_Unimplemented(t *testing.T) {
```
sbalabanov: no need for this test, the real implementation will follow soon. Just remove the test.

controller/getchangedtargetgraph.go
sbalabanov: Updated GetChangedTargetGraph to wrap the Unimplemented YARPC error with NewInfra in 5fc3a487e3df2ac32a4975727e3746228e64845c. This keeps the RPC status explicit for callers while classifying the unimplemented server behavior as an infrastructure failure, including the existing metrics assertions.

[addressed by agent]

controller/getchangedtargetgraph_test.go
sbalabanov: Removed the dedicated stub test in 5fc3a487e3df2ac32a4975727e3746228e64845c. I also removed its BUILD entries and the generated stream mock that was only introduced for that test, while retaining the existing metrics-path coverage for the handler lifecycle.

[addressed by agent]

xytan0056 (approved):

PR #254, opened 2026-07-30
fix(graphrunner): honor requested strategy and reject unknown (audit #6)

sbalabanov:
## Summary

Honor the computation strategy supplied by the upstream request. SHELL requests now select the existing shell graph runner, whose current implementation returns its explicit "not yet implemented" error, while UNSET and NATIVE requests select the native runner. Invalid and unknown strategy values return a user-classified error instead of resolving to a runner.

## Intent

Ensure target-graph computation follows the requested strategy and returns an explicit client-visible error when the strategy cannot be identified.

## Changes

- Preserve UNSET, SHELL, and NATIVE values while mapping the request proto.
- Reject INVALID and out-of-range proto values with `ErrorUser` classification.
- Select `NewShellGraphRunner` for SHELL requests.
- Select `NewNativeGraphRunner` for UNSET and NATIVE requests.
- Return a user-classified error from runner selection for unknown domain values.
- Set the explicit UNSET strategy in in-tree clients and test fixtures that rely on the native default.
- Keep cache keys strategy-specific and retain the existing shell runner stub.

## Test Plan

- [x] `./tools/bazel test //internal/mapper:mapper_test //orchestrator:orchestrator_test //graphrunner:graphrunner_test --test_output=errors`
- [x] Mapper tests cover supported and invalid/unknown strategy values.
- [x] Orchestrator tests cover unknown-strategy error classification.
- [x] `./tools/bazel test //controller:controller_test --test_output=errors`
- [x] `make test-integration`
- [x] `aifx verify --timeout 15`

## Revert Plan

Revert this commit to restore the previous computation-strategy selection behavior.

## Jira Issues

None.

### DIFF
--- a/internal/mapper/build_description.go
+++ b/internal/mapper/build_description.go
@@ -2,7 +2,9 @@ package mapper
 
 import (
 	"errors"
+	"fmt"
 
+	tangoerrors "github.com/uber/tango/core/errors"
 	"github.com/uber/tango/entity"
 	"github.com/uber/tango/tangopb"
 )
@@ -21,11 +23,15 @@ func ProtoToBuildDescription(desc *tangopb.BuildDescription) (entity.BuildDescri
 	if desc.GetBaseSha() == "" {
 		return entity.BuildDescription{}, errors.New("build description base_sha is required")
 	}
+	strategy, err := validateComputationStrategy(desc.GetStrategy())
+	if err != nil {
+		return entity.BuildDescription{}, err
+	}
 	return entity.BuildDescription{
 		Remote:         desc.GetRemote(),
 		BaseSha:        desc.GetBaseSha(),
 		ChangeRequests: toChangeRequests(desc.GetRequests()),
-		Strategy:       toComputationStrategy(desc.GetStrategy()),
+		Strategy:       strategy,
 	}, nil
 }
 
@@ -44,16 +50,21 @@ func toChangeRequests(requests []*tangopb.Request) []entity.ChangeRequest {
 	return out
 }
 
-// toComputationStrategy converts a proto ComputationStrategy to the domain ComputationStrategy.
-func toComputationStrategy(s tangopb.ComputationStrategy) entity.ComputationStrategy {
+// validateComputationStrategy converts a proto ComputationStrategy to the
+// domain type. Supported strategies are preserved so the orchestrator can
+// select the requested runner; invalid or unknown values return a
+// user-classified error.
+func validateComputationStrategy(s tangopb.ComputationStrategy) (entity.ComputationStrategy, error) {
 	switch s {
 	case tangopb.COMPUTATION_STRATEGY_UNSET:
-		return entity.ComputationStrategyUnset
+		return entity.ComputationStrategyUnset, nil
 	case tangopb.COMPUTATION_STRATEGY_SHELL:
-		return entity.ComputationStrategyShell
+		return entity.ComputationStrategyShell, nil
 	case tangopb.COMPUTATION_STRATEGY_NATIVE:
-		return entity.ComputationStrategyNative
+		return entity.ComputationStrategyNative, nil
 	default:
-		return entity.ComputationStrategyInvalid
+		return entity.ComputationStrategyInvalid, tangoerrors.NewUser(
+			fmt.Errorf("unknown computation strategy: %d", int32(s)),
+		)
 	}
 }
--- a/orchestrator/native_orchestrator.go
+++ b/orchestrator/native_orchestrator.go
@@ -27,6 +27,7 @@ import (
 	"github.com/uber/tango/config"
 	"github.com/uber/tango/core/bazel"
 	"github.com/uber/tango/core/cachekey"
+	tangoerrors "github.com/uber/tango/core/errors"
 	"github.com/uber/tango/core/git"
 	"github.com/uber/tango/core/repomanager"
 	"github.com/uber/tango/core/storage"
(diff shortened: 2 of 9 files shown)


sbalabanov (commented):
The desired behavior is, when computation strategy = shell provided by upstream through proto object, the corresponding computation strategy errors out and the error is returned to the client. Shell computation strategy will be implemented later. Analyze if this is already the case, if so, close the PR as "won't do". Otherwise provide a justification and/or a fix and keep PR in draft mode.

sbalabanov:
Thank you for clarifying the intended behavior. I traced the request path on `main` and confirmed that it is not already the case: the proto mapper preserves `COMPUTATION_STRATEGY_SHELL`, but the production orchestrator does not select `NewShellGraphRunner` from that value. Instead, when no runner is injected, it always constructs `NewNativeGraphRunner`; the strategy is only used in cache-key construction. `NewShellGraphRunner` has no production call sites. As a result, a SHELL request can run the native implementation and be cached under a SHELL-labeled key rather than returning the stub's "not yet implemented" error.

The current change at `539c7ac05d2b5f7f320fbd97adde005c0a4f68e1` addresses that gap at the request boundary: SHELL and unknown strategies return a user-classified error before checkout, computation, or cache access, while UNSET continues to default to native behavior. This gives the client the requested error behavior without wiring an intentionally unimplemented runner into production. I also validated the full Bazel build and test suite locally. Per your instruction, I am keeping the PR in draft mode.

[addressed by agent]

sbalabanov (commented):
In this case, rework PR to invoke SHELL computation strategy if requested by the upstream. If requested strategy is unknown, it should be an error rather than silent fallback.

sbalabanov:
Reworked computation strategy selection in 6057b8e0efe8b6f03107cce287fbbe0c6331c4ae.

SHELL requests now construct and invoke NewShellGraphRunner, so its current not-implemented error is returned to the client instead of silently running the native implementation. UNSET and NATIVE continue to use NewNativeGraphRunner, while INVALID and out-of-range values are rejected as user-classified errors before workspace work begins. Focused mapper, orchestrator, and graph-runner Bazel tests pass.

Per the prior instruction, the PR remains in draft mode.

[addressed by agent]

internal/mapper/build_description.go
```diff
@@ -44,16 +50,21 @@ func toChangeRequests(requests []*tangopb.Request) []entity.ChangeRequest {
 	return out
 }
 
-// toComputationStrategy converts a proto ComputationStrategy to the domain ComputationStrategy.
-func toComputationStrategy(s tangopb.ComputationStrategy) entity.ComputationStrategy {
+// validateComputationStrategy converts a proto ComputationStrategy to the
+// domain type. Supported strategies are preserved so the orchestrator can
+// select the requested runner; invalid or unknown values are rejected with a
+// user-classified error rather than silently falling back.
```
sbalabanov: do not use wording that assumes prior behavior, such as "rather than silently falling back"
instead only document current behavior.

internal/mapper/build_description.go
```diff
@@ -44,16 +50,21 @@ func toChangeRequests(requests []*tangopb.Request) []entity.ChangeRequest {
...
+		return entity.ComputationStrategyUnset, nil
 	case tangopb.COMPUTATION_STRATEGY_SHELL:
-		return entity.ComputationStrategyShell
+		return entity.ComputationStrategyShell, nil
 	case tangopb.COMPUTATION_STRATEGY_NATIVE:
-		return entity.ComputationStrategyNative
+		return entity.ComputationStrategyNative, nil
 	default:
-		return entity.ComputationStrategyInvalid
+		return entity.ComputationStrategy(0), tangoerrors.NewUser(
```
sbalabanov: use ComputationStrategyInvalid instead of entity.ComputationStrategy(0)

orchestrator/native_orchestrator.go
```diff
@@ -237,6 +245,17 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, req entity.GetT
 	return graphReader, nil
 }
 
+func validateComputationStrategy(strategy entity.ComputationStrategy) error {
```
sbalabanov: remove this helper function, it is only called from one place. Instead, just error out at where the strategy resolves to a runner if the right strategy cannot be identified.

orchestrator/native_orchestrator_test.go
```diff
@@ -137,6 +137,61 @@ func TestNative_GetTargetGraph_TreehashNotFound_NoError(t *testing.T) {
 	require.NotNil(t, chunk.Targets)
 }
 
+func TestNative_GetTargetGraph_ShellStrategyInvokesShellRunner(t *testing.T) {
```
sbalabanov: change detector test, no need; remove it.

orchestrator/native_orchestrator_test.go
```diff
@@ -137,6 +137,61 @@ func TestNative_GetTargetGraph_TreehashNotFound_NoError(t *testing.T) {
...
+
+	reader, err := o.GetTargetGraph(context.Background(), entity.GetTargetGraphRequest{
+		Build: entity.BuildDescription{
+			Remote:   "git@github:uber/tango",
+			BaseSha:  "1234567890",
+			Strategy: entity.ComputationStrategy(99),
+		},
+	})
+	require.Nil(t, reader)
+	require.EqualError(t, err, "unknown computation strategy: 99")
```
sbalabanov: do not assert on error message, either assert on error type (if applicable), or just on error is not nil

internal/mapper/build_description.go
sbalabanov: Updated the mapper documentation in e293aa856e95976dc106c7244a336eba2cf89d36.

The comment now describes only the current contract: supported strategies are preserved for runner selection, while invalid or unknown values return a user-classified error. It no longer references prior fallback behavior.

[addressed by agent]

orchestrator/native_orchestrator_test.go
sbalabanov: Removed the requested test in e293aa856e95976dc106c7244a336eba2cf89d36.

The focused shell/change-detector test and its dedicated setup are no longer part of the PR. Existing package coverage remains unchanged outside that requested removal.

[addressed by agent]

internal/mapper/build_description.go
sbalabanov: Updated the invalid return value in e293aa856e95976dc106c7244a336eba2cf89d36.

The mapper now returns entity.ComputationStrategyInvalid alongside the user-classified error, making the unsuccessful conversion explicit instead of relying on a numeric zero value.

[addressed by agent]

orchestrator/native_orchestrator_test.go
sbalabanov: Relaxed the assertion in e293aa856e95976dc106c7244a336eba2cf89d36.

The test now checks that an error is returned and that it carries the user error classification, without coupling the test to the exact error message text.

[addressed by agent]

orchestrator/native_orchestrator.go
sbalabanov: Inlined unknown-strategy handling in e293aa856e95976dc106c7244a336eba2cf89d36.

The one-use validation helper was removed, and runner selection now returns the user-classified error directly from its default case when no supported strategy can be identified.

[addressed by agent]

xytan0056 (approved):

PR #260, opened 2026-07-30
docs(storage): clarify stream completion contract (audit #15)

sbalabanov:
## Summary

The storage contract already distinguishes a complete blob from a failed download: the `ReadCloser` returns `io.EOF` only after all blob bytes have been delivered and returns a non-EOF error if the read terminates early. An additional stream header/footer format is therefore unnecessary.

- Clarify the `Storage.Get` completion and termination contract.
- Document that the JSON reader propagates decode and storage read errors.
- Add a test verifying that a terminal storage error is not converted to `io.EOF`.
- Update the compared-target cache comments to describe malformed JSON and failed downloads accurately.

## Test plan

- `./tools/bazel test //core/storage:storage_test //controller:controller_test --test_output=errors`


sbalabanov (commented):
Inspect if the change is really needed. The contract of the storage is to provide an EOF when the end of the stream is reached or another error if it fails to read next record, like stream termination. The client is responsible to properly process the stream that fails mid-flight. If the change is not needed with this new info, update comments in the code accordingly to reflect the contract clearer.

sbalabanov:
Reworked this in c329855141527f994694ae2af054ae18d66e3e14. The completion footer is no longer added; Storage.Get now explicitly requires io.EOF only after the complete blob and a non-EOF error for premature termination. Reader documentation, cache comments, and a test verifying propagation of termination errors were added.

[addressed by agent]

sbalabanov:
useless

PR #261, opened 2026-07-30
fix(controller): compute full BFS before caching compared targets (audit #1)

sbalabanov:
## Summary

`max_distance` from the request's `OutputConfig` was passed into `compareTargetGraphs` / `targetdiff.Compare`, causing the BFS to stop early. The truncated result was then cached under a key that does not include `max_distance`, so a later broader request would replay incomplete distances from the shared cache.

- Always run `targetdiff.Compare` with `MaxDistance: -1` (full BFS) so cached results are complete
- Remove the `maxDist` parameter from `compareTargetGraphs` — it no longer needs it
- Per-request distance filtering continues at send time (`sendTrimmedChangedTargets` and the cache-serve path)
- Add regression test: narrow request (max_distance=1) populates cache, then a broader request (max_distance=-1) on the same treehash pair returns the full target set with correct distances

### DIFF
--- a/controller/getchangedtargets_test.go
+++ b/controller/getchangedtargets_test.go
@@ -139,7 +139,7 @@ func TestCompareTargetGraphs(t *testing.T) {
 	firstGraph := entity.GetTargetGraphResponse{Metadata: &entity.Metadata{}}
 	secondGraph := entity.GetTargetGraphResponse{Metadata: &entity.Metadata{}}
 
-	response, err := c.compareTargetGraphs(t.Context(), c.emitter, zap.NewNop(), []entity.GetTargetGraphResponse{firstGraph}, []entity.GetTargetGraphResponse{secondGraph}, -1)
+	response, err := c.compareTargetGraphs(t.Context(), c.emitter, zap.NewNop(), []entity.GetTargetGraphResponse{firstGraph}, []entity.GetTargetGraphResponse{secondGraph})
 	require.NoError(t, err)
 	require.NotNil(t, response)
 }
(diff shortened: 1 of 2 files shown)


controller/getchangedtargets_test.go
```diff
@@ -1412,6 +1412,177 @@ func TestFetchTargetGraphs(t *testing.T) {
...
+		return nil
+	}).AnyTimes()
+
+	err := c.GetChangedTargets(mkRequest(1), stream1)
+	require.NoError(t, err)
+
+	// Wait for background cache write to complete.
+	select {
+	case <-cacheWriteDone:
+	case <-time.After(2 * time.Second):
```
sbalabanov: do not hardcode timeouts, just wait indefinitely if needed and rely on test runner (i.e. Bazel timeout)

controller/getchangedtargets_test.go
```diff
@@ -1412,6 +1412,177 @@ func TestFetchTargetGraphs(t *testing.T) {
 	})
 }
 
+// TestGetChangedTargets_CacheNotPoisonedByMaxDistance is a regression test for
+// the cache poisoning bug where a narrow max_distance request would cache
+// truncated BFS results (targets beyond the cap left at distance -1), causing
+// a later broader request on the same treehash pair to replay the incomplete
+// result. The fix ensures compareTargetGraphs always runs a full BFS
+// (MaxDistance -1), and per-request filtering happens only at send time.
+func TestGetChangedTargets_CacheNotPoisonedByMaxDistance(t *testing.T) {
```
sbalabanov: no need for regression test, remove it

controller/getchangedtargets_test.go
sbalabanov: Removed the requested regression test in 15cab39ef3f61cf66f6dfb555c99d0e76a7f7b80, which also removes its hardcoded timeout.

The timeout existed only to coordinate the test's background cache write. Since the regression test was removed as requested, no indefinite wait is needed and the production fix remains narrowly covered by the existing controller test suite.

[addressed by agent]

controller/getchangedtargets_test.go
sbalabanov: Removed the regression test in 15cab39ef3f61cf66f6dfb555c99d0e76a7f7b80 while retaining the focused production change and necessary existing test call-site updates.

This keeps the PR scoped to computing the full comparison before cache storage without adding the end-to-end regression fixture the review asked to omit. The affected controller test target passes locally.

[addressed by agent]

xytan0056 (approved):
good catch

PR #263, opened 2026-07-30
fix(cachekey): include config fingerprint in graph keys, skip mapping cache for mutable refs (audit #7)

sbalabanov:
## Summary

Graph cache keys omitted graph-affecting RepositoryConfig fields, so a config change (e.g. enabling bzlmod, adding bazel args) could silently serve stale cached graphs. Additionally, the treehash-mapping cache accepted mutable refs like HEAD, which go stale once the ref moves.

- Add `HashGraphAffectingConfig` to compute an md5 digest of the 7 graph-affecting config fields; emit empty suffix for default/zero config so existing cache entries remain valid
- Extend `GetGraphByTreeHash` with a `configHash` parameter and update all call sites (controller, orchestrator)
- Add `IsFullHexSHA` guard: treehash mapping lookups and writes are skipped for non-40-hex base_sha values (HEAD, branch names, short SHAs)
- Add comprehensive table-driven tests for config hashing (default=empty, each field changes key, order-insensitivity) and mutable-ref skipping
- Update `GetTreehashCachePath` and `ProtoToBuildDescription` godocs to document the mutable-ref constraint

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./core/cachekey/ ./controller/ ./orchestrator/ ./internal/mapper/` -- all pass
- [x] Full `go test ./...` passes (integration tests skip due to missing env, pre-existing)
- [x] `make gazelle` -- BUILD files in sync

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/uber-go/tally"
 	"github.com/uber/tango/config"
+	"github.com/uber/tango/core/cachekey"
 	"github.com/uber/tango/core/storage"
 	"github.com/uber/tango/observability/metrics"
 	"github.com/uber/tango/orchestrator"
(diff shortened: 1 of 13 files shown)


sbalabanov:
questionable approach: validation should happen elsewhere, cached graph might still be right even if some args change.
closing without merge.

PR #264, opened 2026-07-30
feat!: adopt canonical change URIs for change requests

xytan0056:
## Summary

Adopt the [change-URI RFC](https://github.com/uber/submitqueue/blob/main/doc/rfc/change-uri.md) format for change requests. The URI is the self-contained identity of a change: the host names the provider instance, the embedded head SHA pins the code state, and the URI alone determines the cache key. The separate proto `commit` field is removed.

The format is enforced by the bundled native orchestrator, not the gateway — the mapper passes URLs through opaquely, so custom `Orchestrator` implementations can accept formats of their own.

## Accepted URI format

| Provider | Format | Example |
|---|---|---|
| GitHub PR | `github://{host[:port]}/{org}/{repo}/pull/{pr}/{head_sha}` | `github://github.com/uber/tango/pull/123/c3a4b5d6e7f80912a3b4c5d6e7f80912a3b4c5d6` |

`phab://` and `git://` are not supported yet. Violations are rejected, never normalized:

| Component | Rule |
|---|---|
| `host` | required; lowercase only |
| `port` | optional; digits only |
| `org`, `repo` | verbatim, case-sensitive; org may span segments (`uber/frontend`) |
| `pr` | positive integer, no leading zeros |
| `head_sha` | full 40-char lowercase hex |

No userinfo, query, fragment, or empty segments; parse → serialize round-trips byte-for-byte.

## Changes

- New `core/changeuri` package: canonical parser/validator
- Proto: `Request.commit` removed; `tangopb/` regenerated
- Cache key hashes only the URI; native orchestrator rejects invalid URIs as `ErrorUser`
- `workspace.NewRequest` parses the URI and threads PR number + head SHA to `gitRequest`

## Breaking change

The native orchestrator rejects old-format URLs (`github://org/repo/pull/N` + `commit` field); callers must migrate. Old treehash-cache entries recompute; treehash-keyed graph entries are unaffected.

## Test plan
CI
integration test


proto/tango.proto
```diff
@@ -19,12 +19,20 @@ message TangoError {
...
+    // github://{host[:port]}/{org}/{repo}/pull/{pr}/{head_sha}
+    // e.g. github://github.com/uber/tango/pull/123/c3a4b5d6e7f80912a3b4c5d6e7f80912a3b4c5d6
+    // whose embedded head SHA pins the exact code state; custom
+    // orchestrators may accept formats of their own. The URL participates
+    // in cache keys as an opaque string.
     string url = 1;
-    // The commit SHA associated with the change request
-    string commit = 2;
+    // commit was the SHA to apply; it is now embedded in the URI.
+    reserved 2;
```
sbalabanov: as long as we can afford to make breaking changes anyways, we may refrain from reserving a field numeric

proto/tango.proto
```diff
@@ -19,12 +19,20 @@ message TangoError {
...
 message Request {
-    // The URL of the change request (e.g., a GitHub pull request URL or Phabricator diff URL)
+    // The change URI of the request. The format is interpreted by the
+    // Orchestrator implementation: the bundled native orchestrator requires
+    // the canonical change-URI form
+    // github://{host[:port]}/{org}/{repo}/pull/{pr}/{head_sha}
+    // e.g. github://github.com/uber/tango/pull/123/c3a4b5d6e7f80912a3b4c5d6e7f80912a3b4c5d6
+    // whose embedded head SHA pins the exact code state; custom
+    // orchestrators may accept formats of their own. The URL participates
+    // in cache keys as an opaque string.
```
sbalabanov, line 31: important to mention that URI must be immutable to the content it represents, for any possible implementation of the orchestrator

core/changeuri/changeuri.go
```diff
@@ -0,0 +1,174 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package changeuri parses and validates change URIs — the system-wide
+// identity of a code change, per the submitqueue change-URI RFC
+// (https://github.com/uber/submitqueue/blob/main/doc/rfc/change-uri.md).
+//
+// A change URI is an RFC 3986 URI of the form scheme://{host[:port]}/{path}
+// whose path pins the change to an exact code state. Only the GitHub pull
```
sbalabanov, line 20: to make future implementations more structured, should we place github implementation under changeuri/github folder? Or even adopt the package from submitqueue?

core/changeuri/changeuri.go
```diff
@@ -0,0 +1,174 @@
...
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// Scheme is the URI scheme for GitHub pull requests.
+const Scheme = "github"
+
+// PullRequest is the parsed form of a canonical GitHub pull request URI.
```
sbalabanov, line 41: the "canonical" Github only has PR number, not a hash I think?

core/changeuri/changeuri.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	if u.User != nil {
+		return PullRequest{}, fmt.Errorf("change URI must not contain userinfo: %q", raw)
+	}
+	if u.RawQuery != "" || u.ForceQuery || u.Fragment != "" {
+		return PullRequest{}, fmt.Errorf("change URI must not contain a query or fragment: %q", raw)
+	}
+	if err := validateHost(u); err != nil {
+		return PullRequest{}, err
+	}
```
yushan8, line 77: nit: should these lines be in a helper function `validateURL`. I think it's a little bit more readable that way so this function focuses on parsing the pull request string into the struct.

xytan0056: Done

core/changeuri/changeuri.go
```diff
@@ -0,0 +1,181 @@
+// Copyright (c) 2025 Uber Technologies, Inc.
```
yushan8:
```suggestion
// Copyright (c) 2026 Uber Technologies, Inc.
```

proto/tango.proto
```diff
@@ -19,12 +19,20 @@ message TangoError {
...
     string url = 1;
-    // The commit SHA associated with the change request
-    string commit = 2;
+    // commit was the SHA to apply; it is now embedded in the URI.
+    reserved 2;
+    reserved "commit";
```
yushan8: should we just remove it?

yushan8 (approved):

core/changeuri/changeuri.go
xytan0056: Done in f4264d9.

proto/tango.proto
xytan0056: once remove and deploy the package, existing calls can fail

yushan8: I think we should just remove it for now to keep the proto clean. This isn't used by anyone but us, I think we can update it until it's actually finalized and productionized. Once it's actually in production, I agree we should use that to deprecated the fields.

PR #266, opened 2026-07-31
perf(orchestrator): store treehash mapping before graph computation

xytan0056:
## Summary
- Moves the treehash cache write (`BuildDescription → treehash`) from after graph computation to before it, so concurrent/subsequent requests can resolve the mapping while the current request is still computing the target graph.
- Makes the write fire-and-forget: failures are logged as warnings but don't fail the request, since the mapping is a performance optimization, not a correctness requirement.
- The treehash mapping is a few bytes, so the synchronous `Put` is negligible compared to graph compute — the win is in timing, not concurrency.

## Test plan
- [x] `bazel test //orchestrator/...` passes
- [x] `make build` passes

### DIFF
--- a/orchestrator/native_orchestrator.go
+++ b/orchestrator/native_orchestrator.go
@@ -186,6 +186,26 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, req entity.GetT
 	} else {
 		logger.Infow("GetTargetGraph: bypass_cache=true, computing target graph")
 	}
+	// Store the treehash mapping in the background before the (potentially
+	// slow) graph computation so concurrent or subsequent requests can
+	// resolve it without waiting for the graph to finish.
+	go func() {
+		bgOp := metrics.Begin(e, _opTreehashCacheWrite, metrics.FastDurationBuckets)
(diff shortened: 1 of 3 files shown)


yushan8 (approved):

orchestrator/native_orchestrator.go
```diff
@@ -186,6 +186,20 @@ func (b *nativeOrchestrator) GetTargetGraph(ctx context.Context, req entity.GetT
...
 	}
+	// Store the treehash mapping in the background before the (potentially
+	// slow) graph computation so concurrent or subsequent requests can
+	// resolve it without waiting for the graph to finish.
+	go func() {
+		thCachePath := cachekey.GetTreehashCachePath(build)
+		if putErr := b.storage.Put(b.appCtx, storage.UploadRequest{
+			Key:    thCachePath,
+			Reader: bytes.NewReader([]byte(treehash)),
+		}); putErr != nil {
```
yushan8: shouldn't we return the error if the treehash upload fails? the request will fail if there's no commit -> treehash mapping?

yushan8: synced offline: Add a metric to track when treehash mapping upload failed

PR #269, opened 2026-08-03
fix(targethasher): make directory hashing cancellation-aware (audit #14.3)

sbalabanov:
## Summary

**Audit #14.3.** Propagate cancellation into source-directory hashing across both target graph implementations.

## Intent

Stop expensive directory traversal when graph computation is canceled and preserve custom cancellation causes through the hashing boundary.

## Changes

- Add `context.Context` to the `SourceHasher` contract and update all implementations, callers, fakes, and mocks.
- Check cancellation before directory traversal and periodically while hashing files.
- Preserve `context.Cause` and verify both graph paths pass the original context.

## Test Plan

- [x] `go test ./core/targethasher ./core/itg/graph`
- [x] `./tools/bazel test //core/targethasher:targethasher_test //core/itg/graph:graph_test --test_output=errors`
- [x] `make gazelle`
- [x] `aifx verify`

## Revert Plan

Revert this PR to restore the context-free source hashing interface and traversal behavior.

## Jira Issues

None.

### DIFF
--- a/core/targethasher/sourcehasher.go
+++ b/core/targethasher/sourcehasher.go
@@ -15,6 +15,7 @@
 package targethasher
 
 import (
+	"context"
 	"crypto/sha1"
 	"fmt"
 	"hash"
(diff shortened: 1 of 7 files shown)


core/targethasher/sourcehasher.go
```diff
@@ -37,11 +38,16 @@ const (
 	_defaultSourceFileVisibility = "//visibility:private"
 )
 
+// cancelCheckInterval is how often the directory walk checks for cancellation,
+// measured in files hashed. Matches the convention used elsewhere in the
```
sbalabanov: Do not use words "matches the convention", just describe what this setting is for.

sbalabanov: Updated the comment in 35d2ee5cc4bddfdd1d99b88ad80cbc8e37293d67 to directly state that the setting controls how many files are hashed between cancellation checks during a directory walk.

[addressed by agent]

xytan0056 (approved):

PR #271, opened 2026-08-06
fix(workspace): fetch PR refs from real remote, not local origin

xytan0056:
## Summary

- Worker clones created by the repo manager use `git clone --local`, so their `origin` remote points to the local origin directory — which has no `refs/pull/*` refs. `gitRequest.Apply()` was fetching from `"origin"`, causing `couldn't find remote ref pull/NNN/head` errors.
- Thread the real remote URL (`build.Remote`) through `NewRequest` and `NewGitRequest` so the fetch targets the actual GitHub remote.
- Also fix path parsing to properly extract the PR number from both legacy (`github://org/repo/pull/123`) and canonical (`github://host/org/repo/pull/123/head_sha`) URI formats.

## Test plan
CI

Local test
```
make run-server

➜  tango bazel run //example/client -- \
  -method get-changed-targets \
  -remote https://github.com/uber/tango.git \
  -base-sha 6eaa518b66f0fa7e6822cef021df9618ff8fb10e \
  -new-base-sha 6eaa518b66f0fa7e6822cef021df9618ff8fb10e \
  -new-request-urls 'github://github.com/uber/tango/pull/191/2224a38ca1d3cd01b38e1102e6ec6a2baed1becb'
```

### DIFF
--- a/core/workspace/request.go
+++ b/core/workspace/request.go
@@ -31,7 +31,7 @@ type Request interface {
 
 // NewRequest creates a new request from a canonical change URI, e.g.
 // github://{host[:port]}/{org}/{repo}/pull/{pr}/{head_sha}.
-func NewRequest(rawURL string, g git.Interface, baseRef string, logger *zap.SugaredLogger) (Request, error) {
+func NewRequest(rawURL string, g git.Interface, remote string, baseRef string, logger *zap.SugaredLogger) (Request, error) {
 	u, err := url.Parse(rawURL)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


core/workspace/request.go
```diff
@@ -31,18 +31,37 @@ type Request interface {
...
 	}
 	switch u.Scheme {
-	case changeuri.Scheme:
-		pr, err := changeuri.Parse(rawURL)
-		if err != nil {
-			return nil, fmt.Errorf("invalid change URI %q: %w", rawURL, err)
+	case "github":
+		prNumber, headSHA := parseGitHubPath(u.Path)
+		if prNumber == "" {
+			return nil, fmt.Errorf("could not extract PR number from path %q", u.Path)
```
yushan8: nit: we should probably classify these as `UserError` on the orchestrator.

core/workspace/request.go
```diff
@@ -31,18 +31,37 @@ type Request interface {
...
 		}
-		return NewGitRequest(g, pr.Number, baseRef, pr.HeadSHA, logger), nil
+		if headSHA == "" {
+			return nil, fmt.Errorf("change URI must include a head SHA: %q", rawURL)
```
yushan8: same here

yushan8 (approved):

PR #273, opened 2026-08-06
[feature] Add per-repo allowlist for direct-change attribute detection

yushan8:
## Summary
Intent:
- GetChangedTargets classifies a target as directly changed (distance 0)
  whenever its attribute set differs between revisions.
- Some Bazel attributes are pure BUILD-file bookkeeping (e.g.
  generator_location, which shifts whenever unrelated lines move earlier in
  the same file) rather than semantic configuration, so comparing the full
  attribute set can misclassify cosmetic churn as a direct change.

Changes:
- Add RepositoryConfig.DirectlyChangedAttributes, a per-repo allowlist of
  attribute names that count as evidence of a direct change.
- Wire config.RepositoryConfigProvider into controller.Params/controller so
  GetChangedTargets can look up the allowlist for a request's remote.
- Thread the resolved allowlist through compareTargetGraphs and toDiffGraph
  so only allowlisted attributes (or all attributes, when unset) are kept on
  each target before diffing.

## Test Plan
- Added TestToDiffGraph_OnlyKeepsAllowlistedAttributes and
  TestDirectlyChangedAttributesFor covering the allowlist filtering and
  lookup logic.
- Updated existing compareTargetGraphs/toDiffGraph call sites for the new
  parameter.

### DIFF
--- a/config/repository_config.go
+++ b/config/repository_config.go
@@ -32,7 +32,7 @@ type RepositoryConfig struct {
 	// BazelStartupOptions are Bazel startup flags placed before the `query` subcommand (e.g. "--batch"); empty by default.
 	BazelStartupOptions []string `yaml:"bazel_startup_options"`
 	StreamBazelLogs     bool     `yaml:"stream_bazel_logs"`
-	// DirectlyChangedAttributes restricts which rule attributes GetChangedTargets
+	// SeedAttributes restricts which rule attributes GetChangedTargets
 	// treats as evidence that a target's own configuration changed, as opposed to
 	// dependency-only churn.
 	//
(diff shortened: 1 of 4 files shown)


config/repository_config.go
```diff
@@ -26,12 +22,35 @@ type RepositoryConfig struct {
...
+	// BUILD file, with no effect on the rule's behavior. Comparing the full,
+	// unfiltered attribute set would classify that cosmetic churn as a direct
+	// change and assign the target distance 0, when it should instead inherit
+	// its distance from its dependencies.
+	//
+	// When DirectlyChangedAttributes is set, only attributes named here are
+	// compared; all others are ignored for this classification. When unset
+	// (the default), every attribute is compared and can trigger a direct-change
+	// classification.
+	DirectlyChangedAttributes []string `yaml:"directly_changed_attributes"`
```
xytan0056: The naming here could be more accurate, it reads unclear. How about`SeedAttributes`?

xytan0056 (approved):

PR #274, opened 2026-08-06
fix(controller): use a custom cancel cause for canceling graph fetch jobs

justinwon777:
Cancelling a job's context on sibling failure now carries a distinct cause instead of the default context.Canceled, so it's clear the cancellation was self-inflicted rather than a client disconnect.

## Test Plan

unit test

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -42,7 +42,7 @@ type job struct {
 	cancelled   bool
 	completed   bool
 	ctx         context.Context
-	cancel      context.CancelFunc
+	cancel      context.CancelCauseFunc
 }
 
 // GetChangedTargets returns the changed targets between two revisions. If the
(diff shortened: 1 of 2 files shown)


controller/getchangedtargets.go
```diff
@@ -289,7 +289,7 @@ func (c *controller) fetchTargetGraphs(ctx context.Context, e *metrics.Emitter,
 		if jobs[res.order].err != nil {
 			other := (res.order + 1) % 2
 			if !jobs[other].completed {
-				jobs[other].cancel()
+				jobs[other].cancel(errors.New("sibling graph fetch failed"))
```
xytan0056, line 292: should this be tangoerrors.NewUser or Infra?

justinwon777: we ignore the error from canceled job, so there's no need to classify it.

yushan8 (approved):

controller/getchangedtargets.go
sywhang: sorry if this is a naive question, but why?

justinwon777: When a job has an error, we cancel the other job and return the first job's err. Whatever happens to the canceled job doesn't matter since we're intentionally canceling it.

PR #280, opened 2026-08-11
TGB (tango graph binary) implementation

sywhang:
This implements tgb format described in #279 as a drop-in replacement. The existing APIs did not change at all, and enrollment can be optionally configured using service-level config knobs.


xytan0056 (approved):

sywhang:
i feel bad for merging a 8K line PR at once... but... yolo.

PR #292, opened 2026-08-26
fix(controller): detect AllTargetsFiles deletion (BUG-013)

sbalabanov:
## Summary

Fix BUG-013 by detecting when a configured `AllTargetsFiles` path is added or deleted between revisions.

Previously, Tango compared only files present in each revision, so deleting a global trigger file could be missed. For example, if `.bazelrc` is configured as an `AllTargetsFiles` entry and is removed in a new revision, every target must be invalidated because the repository-wide build configuration changed.

The fix records configured paths even when absent and compares the complete old and new path/hash sets symmetrically.

## Test Plan

- Unit tests cover added, deleted, missing-side, and unchanged configured files.

## Issues

T3-BUG-013

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"time"
 
 	"github.com/uber/tango/core/cachekey"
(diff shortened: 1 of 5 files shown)


sbalabanov (commented):
Simplify PR description, add use case example of the bug it is fixing

sbalabanov:
Simplified the PR description and added a concrete `.bazelrc` deletion example showing why removal of an AllTargetsFiles trigger must invalidate every target. Current HEAD: `7c7c5c4e74f43f7fb2cb538f7fb212f8597f6c51`.

[addressed by agent]

xytan0056 (approved):
good catch

xytan0056 (approved):

PR #294, opened 2026-08-26
fix(api)!: remove roots_only output option (BUG-021)

sbalabanov:
## Summary

Fix BUG-021 by removing the unsupported `OutputConfig.roots_only` option from the Tango API and regenerated Go/YARPC bindings.

Because Tango has not reached a major release, this intentionally makes the breaking protobuf change directly without reserving the removed field name or number.

## Test Plan

- `PATH=/tmp/tango-proto-tools/bin:$PATH make proto` (run twice; deterministic output confirmed)
- `go test ./controller ./mapper ./internal/mapper ./tangopb/...`
- `./tools/bazel build //proto:tangopb_go_proto //tangopb:tangopb`
- `./tools/bazel test //controller:controller_test --test_output=errors`

## Breaking Change

`OutputConfig.roots_only` is removed from the protobuf contract and generated Go/YARPC bindings.


sbalabanov (commented):
Instead just remove roots_only from the API contract. Do not bother with backwards compatibility.

sbalabanov:
Addressed the review in 5c72105dea702e8bc4450bc7c2c56efac7d9f7a0 by removing OutputConfig.roots_only from the proto contract and regenerating the Go and YARPC bindings. The prior controller filtering implementation and tests were dropped, so BUG-021 now performs the requested contract removal without a compatibility shim.

[addressed by agent]

proto/tango.proto
```diff
@@ -71,8 +71,6 @@ message OutputConfig {
     bool include_attributes = 2;
     // Indicates if hashes should be computed for each target and included in the response
     bool include_hashes = 3;
-    // Indicates if only root targets (i.e. targets that are not depended on by any other targets) should be included in the response
-    bool roots_only = 4;
```
sbalabanov, line 75: reserve ordinal and name

sbalabanov: Reserved OutputConfig field number 4 and field name roots_only in 2e49c71631de60c8c5316c03612d731b718d0744 and regenerated the committed Go/YARPC bindings.

This prevents reuse of the removed field wire ordinal or schema name while preserving the intended API removal.

[addressed by agent]

proto/tango.proto
```diff
@@ -65,14 +65,15 @@ enum ComputationStrategy {
 // OutputConfig controls what payload details are included in responses.
 // Some flags may be ignored depending on the specific API.
 message OutputConfig {
+    reserved 4;
+    reserved "roots_only";
```
xytan0056:
```suggestion
```
we're not on major release yet. we can afford to make breaking changes

sbalabanov: ok, let's change `max_distance` ordinal from 5 to 4 then

sbalabanov: also, update PR description accordingly

sbalabanov: Removed both `roots_only` reservations and regenerated the protobuf bindings in `b3a1d615501e03a31904e68f1ea23936108a75fd`. Since Tango is pre-1.0, the field name and number remain available rather than carrying compatibility reservations.

[addressed by agent]

xytan0056 (approved):

PR #295, opened 2026-08-26
fix(targethasher): frame rule hashes for BUG-006

sbalabanov:
## Summary

Fix BUG-006 by type- and length-framing rule data before hashing.

Example: changing a string-list attribute from `["a", "bc"]` to `["ab", "c"]` previously contributed the same raw bytes (`abc`) to the hash, so Tango could miss a changed target. Canonical serialization makes these values distinct while retaining deterministic Bazel attribute and collection ordering.


core/targethasher/sourcehasher.go
```diff
@@ -37,6 +38,7 @@ var newHash = sha1.New
 const (
 	externalWorkspaceFilePrefix  = "@"
 	_defaultSourceFileVisibility = "//visibility:private"
+	ruleHashEncodingVersion      = "tango.rule.v1"
```
sbalabanov: do not version

sbalabanov: Removed the standalone rule-hash encoding version marker in da6e74c81a36a952ee5d130f42b6f34cdf938c3f.

The canonical encoder still frames field names, value types, presence, lengths, collection counts, and nested messages, so BUG-006 collision resistance remains intact without embedding a version. I updated only the affected golden hashes and did not include BUG-007 behavior.

[addressed by agent]

core/targethasher/sourcehasher.go
```diff
@@ -192,196 +193,287 @@ func externalTargetForRule(t string) string {
...
+	_, _ = h.Write(encoder.buffer.Bytes())
+}
+
+func encodeRule(r *buildpb.Rule) []byte {
+	return encodeMessage("Rule", func(encoder *canonicalEncoder) {
+		encoder.writeOptionalString("name", r.Name)
+		encoder.writeOptionalString("rule_class", r.RuleClass)
+
+		// Location is machine-local and intentionally excluded.
+		attributes := make([]canonicalElement, 0, len(r.GetAttribute()))
```
sbalabanov: save r.GetAttribute() to a variable first to avoid potential expensive function call

core/targethasher/sourcehasher.go
```diff
@@ -192,196 +193,287 @@ func externalTargetForRule(t string) string {
...
+	canonicalMessage canonicalValueType = iota + 1
+	canonicalString
+	canonicalBool
+	canonicalInt32
+	canonicalEnum
+	canonicalList
+)
+
+type canonicalElement struct {
+	present bool
```
sbalabanov: no need to check presence. If some data structure in pb is nil, it can be simply ignored.

sbalabanov (commented):
Shorten diff description and list edge cases in serialization it addresses

core/targethasher/sourcehasher.go
sbalabanov: Cached the attribute slice once and reused it for capacity sizing and iteration in ba296a4bab4df58aef8ae15459912ad4aabfc67c. This avoids duplicate getter evaluation while leaving attribute filtering, sorting, and canonical payload bytes unchanged.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Removed per-element presence tracking and now skip nil protobuf messages in ba296a4bab4df58aef8ae15459912ad4aabfc67c. Non-nil list framing remains unchanged, and regression coverage verifies that nil top-level and nested messages do not affect the rule hash.

[addressed by agent]

sbalabanov:
Shortened the PR description and listed the BUG-006 serialization edge cases in ba296a4bab4df58aef8ae15459912ad4aabfc67c. The concise body now covers boundary, type/presence, ordering, nested-message, and nil-message cases.

[addressed by agent]

core/targethasher/sourcehasher.go
```diff
@@ -192,196 +193,284 @@ func externalTargetForRule(t string) string {
...
+const (
+	canonicalMessage canonicalValueType = iota + 1
+	canonicalString
+	canonicalBool
+	canonicalInt32
+	canonicalEnum
+	canonicalList
+)
+
+type canonicalElement struct {
```
sbalabanov: check if a structure is required as it only contains one type. If needed, add a code comment to explain.

core/targethasher/sourcehasher.go
```diff
@@ -192,196 +193,284 @@ func externalTargetForRule(t string) string {
...
+			// generator_location identifies a macro call site and does not affect
+			// the rule's behavior.
+			if attribute.GetName() == "generator_location" {
+				continue
+			}
+			attributes = append(attributes, canonicalElement{
+				payload: encodeAttribute(attribute),
+			})
+		}
+		sortCanonicalElements(attributes)
```
sbalabanov: check and explain why older sorting by name is insufficient. If sufficient, restore to base algorithm.

core/targethasher/sourcehasher.go
sbalabanov: Removed the one-field canonicalElement wrapper and now pass canonical payloads directly as [][]byte in 3f6e0929847d1d0bd3d68e7fe916920e789921ca. Sorting and list framing remain byte-for-byte unchanged, while the unnecessary structure is eliminated.

[addressed by agent]

sbalabanov: The canonicalElement wrapper removal remains in the final amended BUG-006 commit d5efe0ccdb7041aa600497b14c510c7cc048ec23. This SHA supersedes the earlier reply after the separate attribute-ordering feedback was incorporated; list sorting and framing are still unchanged.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Kept full canonical-byte sorting in d5efe0ccdb7041aa600497b14c510c7cc048ec23 because name-only sorting does not total-order distinct attributes with the same name. The protobuf schema does not require names to be unique, so reversed duplicate-name inputs could otherwise hash differently; a code comment and regression test now document this determinism case.

[addressed by agent]

sbalabanov (commented):
Update PR description with the example of the use case of the bug the PR is fixing

core/targethasher/sourcehasher.go
```diff
@@ -192,196 +193,278 @@ func externalTargetForRule(t string) string {
...
+	_, _ = e.buffer.Write(encoded[:length])
+}
+
+func encodeInt32(value int32) []byte {
+	var encoded [4]byte
+	binary.BigEndian.PutUint32(encoded[:], uint32(value))
+	return encoded[:]
+}
+
+func sortCanonicalElements(elements [][]byte) {
```
sbalabanov: sortCanonicalElements is a bad name now, come up with a better one like sortElements()

core/targethasher/sourcehasher.go
sbalabanov: Bazel ensures there are no duplicate attributes. Restore to sorting by name and add a comment explaining this. Check my statement across Bazel knowledge base first.

sbalabanov:
Updated and simplified the PR description with a concrete BUG-006 failure example in baaabf99a4453dde8f1ba04a43e1538345a4f490. It now shows how ["a", "bc"] and ["ab", "c"] previously serialized to the same bytes and could cause changed-target detection to miss a real update.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Renamed sortCanonicalElements to sortElements in baaabf99a4453dde8f1ba04a43e1538345a4f490. The helper now has a concise name matching its role of ordering encoded byte-slice elements.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Restored attribute sorting by name in baaabf99a4453dde8f1ba04a43e1538345a4f490 after confirming Bazel rule classes enforce one attribute per name. Added a concise comment documenting that invariant, removed the invalid duplicate-name regression, and updated the affected hash snapshots.

[addressed by agent]

xytan0056 (approved):

PR #297, opened 2026-08-26
fix(bazel): coordinate query pipes for BUG-024

sbalabanov:
### Intent
- Fix BUG-024 by preventing Bazel queries from hanging when descendants retain output descriptors.
- Use case: a Bazel query starts a helper process that inherits stdout/stderr and outlives Bazel; Tango then waits indefinitely for EOF instead of returning or honoring cancellation.

### Changes
- Route child output through managed bridge pipes so `Wait` runs while readers drain and `WaitDelay` can close retained descriptors.
- Preserve cancellation causes and join all stream readers after command termination.
- Add deterministic subprocess regressions for descriptor retention and cancellation.

## Test Plan
- `go test -race ./core/bazel`
- `go test ./core/bazel -run 'TestExecuteQueryInternal_(WaitRunsWhileStreamsDrain|DescriptorRetentionDoesNotDeadlock|ContextCancellationTerminatesProcess)$' -count=20`
- `./tools/bazel test //core/bazel:bazel_test`
- `make build`
- `make test`
- `aifx verify`

## Revert Plan
- Revert this PR to restore the previous Bazel subprocess pipe coordination.

---

<sub>Generated by the 🪄 pr-create skill in devexp-agent-marketplace</sub>

### DIFF
--- a/core/bazel/BUILD.bazel
+++ b/core/bazel/BUILD.bazel
@@ -29,6 +29,7 @@ go_test(
     embed = [":bazel"],
     deps = [
         "//core/bazel/commandermock",
+        "//core/execcmd",
         "@com_github_bazelbuild_buildtools//build_proto",
         "@com_github_stretchr_testify//assert",
         "@com_github_stretchr_testify//require",
--- a/core/bazel/bazel.go
+++ b/core/bazel/bazel.go
@@ -86,7 +86,7 @@ func NewBazelClient(ctx context.Context, p Params) (*BazelClient, error) {
 				cmd.Env = append(cmd.Env, key+"="+value)
 			}
 			cmd.Stdin = nil
-			return cmd
+			return &execCommander{Cmd: cmd}
 		}
 	}
 	timeout := p.QueryTimeout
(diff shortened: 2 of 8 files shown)


sbalabanov (commented):
Update PR description with the use case example of the bug that PR addresses

sbalabanov:
Updated the PR description with a concrete BUG-024 use case: a Bazel query helper can outlive Bazel while retaining inherited stdout/stderr, causing Tango to wait indefinitely for EOF instead of returning or honoring cancellation. The final head remains f946992087c741488ad1a07b991442296ba2cfde; this was a metadata-only change with no code push.

[addressed by agent]

sbalabanov (commented):
Swithing to io.Pipe() might have other unintended consequences. Given that bazel query generally behaves, do not attempt to fix. Abandon the diff and close the PR.

sbalabanov:
Acknowledged. Per the latest review, this approach is being abandoned and PR #297 is being closed without further code changes.

[addressed by agent]

PR #298, opened 2026-08-26
fix(controller): preserve AllTargetsFiles change types for BUG-015

sbalabanov:
## Summary

Fix BUG-015 by preserving `NEW` and `DELETED` change types when an `AllTargetsFiles` global trigger changes.

Previously, global invalidation could overwrite every result as `CHANGED`. For example, if `.bazelrc` changes in the same revision that adds `//service:new` and deletes `//service:old`, callers must still receive `NEW` for the added target and `DELETED` for the removed target; only targets present in both revisions should be promoted to distance-zero `CHANGED`.

The fix applies the same classification behavior to chunk-backed and TGB-backed graph comparisons.

## Test Plan

- Unit tests cover unchanged, hash-changed, added, and deleted targets for both graph formats.

## Issues

T3-BUG-015

### DIFF
--- a/controller/getchangedtargets.go
+++ b/controller/getchangedtargets.go
@@ -428,7 +428,7 @@ func (c *controller) compareFetchedGraphs(ctx context.Context, e *metrics.Emitte
 		) {
 			logger.Info("compareFetchedGraphs: AllTargetsFiles trigger matched (TGB), reporting all targets as changed")
 			e.Counter(opGetChangedTargets, "all_targets_triggered").Inc(1)
-			return c.allTargetsChangedFromTGB(ctx, second.tgb.TGB())
+			return c.allTargetsChangedFromTGB(ctx, first.tgb.TGB(), second.tgb.TGB())
 		}
 		return c.compareTargetGraphsTGB(ctx, e, logger, first.tgb.TGB(), second.tgb.TGB(), seedAttrs)
 	}
@@ -563,7 +563,7 @@ func (c *controller) compareTargetGraphs(ctx context.Context, e *metrics.Emitter
 	if allTargetsFileChanged(firstMetadata, secondMetadata) {
 		logger.Info("compareTargetGraphs: AllTargetsFiles trigger matched, reporting all targets as changed")
 		e.Counter(opGetChangedTargets, "all_targets_triggered").Inc(1)
-		return c.allTargetsChangedFromGraph(ctx, secondTargetsByID, secondMetadata)
+		return c.allTargetsChangedFromGraph(ctx, firstTargetsByID, firstMetadata, secondTargetsByID, secondMetadata)
 	}
 
 	before, err := toDiffGraph(ctx, firstTargetsByID, firstMetadata, seedAttrs)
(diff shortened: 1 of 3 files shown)


sbalabanov (commented):
Simplify the PR description, include use case example of the bug the PR fixes

sbalabanov:
Simplified the PR description and added a concrete `.bazelrc` example showing that a global trigger must preserve `NEW` and `DELETED` classifications while promoting only shared targets to `CHANGED`. Current HEAD: `1eeb5a17a9e527f6db6ad36ea5a67b60b10726fc`.

[addressed by agent]

controller/getchangedtargets.go
```diff
@@ -718,38 +718,106 @@ func allTargetsFileChanged(first, second *entity.Metadata) bool {
...
+		return nil, err
+	}
+	return c.resultToResponses(result)
+}
+
+// allTargetsChanged retains the ordinary comparison's membership changes and
+// promotes every target present in both revisions to a distance-zero CHANGED
+// result because the global input change may affect it independently of its
+// target hash.
+func allTargetsChanged(ctx context.Context, before, after targetdiff.Graph) (targetdiff.Result, error) {
```
sbalabanov, line 750: check if implementation has to do two passes over targets "before" and "after". If so, change it to single pass or explain why you can't. I envision iterating over "before" and fill changed targets and deleted targets, then iterate over "after" and fill new targets. No intermediate data structure, like the one returned from `targetdiff.Compare`, is necessary.

sbalabanov: Replaced the targetdiff.Compare call with two direct graph passes in d307022a3ebd59a8545d4109e51e4b426a8ef491. The first pass over before emits CHANGED for shared targets and DELETED for missing targets; the second pass over after emits only NEW targets. This preserves the AllTargetsFiles semantics and distance 0 while avoiding the intermediate comparison, seed, and distance data structures.

[addressed by agent]

xytan0056 (commented):
could you add an integration test for such behavior in 
https://github.com/xytan0056/bazel-fixture/pulls
and expand the testcase [here](https://github.com/uber/tango/blob/374b5a954008894d52dab4d347a8ed678003b177/Makefile#L18-L21) ?

sbalabanov:
> could you add an integration test for such behavior in https://github.com/xytan0056/bazel-fixture/pulls and expand the testcase [here](https://github.com/uber/tango/blob/374b5a954008894d52dab4d347a8ed678003b177/Makefile#L18-L21) ?

why would the fixture go to a private fork?
I could add an integration test right in tango repo if needed, even though the change has a controller test already.

Also for e2e test, it might be a separate diff if we did not test all-in files before.

xytan0056:
> > could you add an integration test for such behavior in https://github.com/xytan0056/bazel-fixture/pulls and expand the testcase [here](https://github.com/uber/tango/blob/374b5a954008894d52dab4d347a8ed678003b177/Makefile#L18-L21) ?
> 
> why would the fixture go to a private fork? I could add an integration test right in tango repo if needed, even though the change has a controller test already.
> 
> Also for e2e test, it might be a separate diff if we did not test all-in files before.

I think it's a true e2e test to run a separate repo outside of tango with PR parsing etc. So that we can test against UX -- actual PR that changes any all_targets_file, rather than hardcoding values like `"old-hash"` `"new-hash"`.

If we test within a directory in tango, the PR's base will have to keep up as the repo evolves.

xytan0056 (approved):

PR #301, opened 2026-08-26
fix(targethasher): frame directory hashes for BUG-007

sbalabanov:
## Summary

BUG-007 fixes directory hashes for edge cases that file contents alone miss:

- A file, directory, or symlink rename changes the hash.
- Directory nesting and empty directories change the hash.
- Adding a symlink or changing its target changes the hash.
- Creation order does not change the hash.

The implementation streams `filepath.WalkDir` in lexical order and hashes each entry's local name, kind, and file contents or symlink target. It does not hash checkout paths or filesystem mode bits, so the result stays stable across machines.

## Tests

Added coverage for rename, directory structure, symlinks, creation order, and mode independence.

### DIFF
--- a/core/targethasher/sourcehasher.go
+++ b/core/targethasher/sourcehasher.go
@@ -40,9 +40,15 @@ const (
 	_defaultSourceFileVisibility = "//visibility:private"
 )
 
-// cancelCheckInterval is the number of files hashed between cancellation
-// checks during a directory walk.
+// cancelCheckInterval is the number of directory entries processed between
+// cancellation checks during a directory walk.
 const cancelCheckInterval = 1024
+
+const (
+	directoryEntryMarker byte = iota
+	regularFileEntryMarker
+	symlinkEntryMarker
+)
 
 // SourceHasher provides hashes for source nodes in the target graph. These
 // can be calculated based on disk contents or form other sources such as a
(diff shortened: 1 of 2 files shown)


core/targethasher/sourcehasher.go
```diff
@@ -37,10 +38,11 @@ var newHash = sha1.New
 const (
 	externalWorkspaceFilePrefix  = "@"
 	_defaultSourceFileVisibility = "//visibility:private"
+	directoryHashEncodingVersion = "tango.directory.v1"
```
sbalabanov: Don't version, backwards compatibility is not required

core/targethasher/sourcehasher.go
```diff
@@ -150,31 +152,135 @@ func hashDir(ctx context.Context, root string) (hash.Hash, error) {
...
 		}
-		return nil
 	}
 
-	err := filepath.WalkDir(root, walkDirFunc)
-	return dirHash, err
+	return dirHash, nil
+}
+
+func relevantDirectoryMode(mode fs.FileMode) uint32 {
```
sbalabanov: do not make it a part of the hash. Make a comment in the appropriate place that mode is omitted.

core/targethasher/sourcehasher.go
```diff
@@ -150,31 +152,135 @@ func hashDir(ctx context.Context, root string) (hash.Hash, error) {
...
+		return err
+	}
+	if !info.Mode().IsRegular() {
+		return fmt.Errorf("%q is no longer a regular file", path)
+	}
+	if info.Size() < 0 {
+		return fmt.Errorf("%q has invalid size %d", path, info.Size())
+	}
+
+	writeDirectoryUint64(h, uint64(info.Size()))
```
sbalabanov: check if this is really needed - shouldn't size be covered by content hash anyways?

core/targethasher/sourcehasher.go
```diff
@@ -150,31 +152,135 @@ func hashDir(ctx context.Context, root string) (hash.Hash, error) {
...
+				return err
+			}
+		}
+		entryCount++
+
+		relativePath, err := filepath.Rel(root, path)
+		if err != nil {
+			return err
+		}
+		entries = append(entries, directoryEntry{
```
sbalabanov: avoid intermediate data structures as walking the monorepo might be memory-intense. Instead, hash as you walk - as before. If filepath.WalkDir is not deterministic - change to another function or iterate manually (leave a comment why walkdir is not used in such a case).

core/targethasher/sourcehasher.go
```diff
@@ -150,31 +152,135 @@ func hashDir(ctx context.Context, root string) (hash.Hash, error) {
...
+			}
+		}
+
+		info, err := os.Lstat(entry.path)
+		if err != nil {
+			return nil, err
+		}
+
+		writeFramedDirectoryBytes(dirHash, []byte(entry.relativePath))
+		writeDirectoryUint32(dirHash, uint32(info.Mode().Type()))
```
sbalabanov: make sure this is platform-agnostic, If needed, only support well-known modes and error otherwise.

core/targethasher/sourcehasher.go
```diff
@@ -150,31 +152,135 @@ func hashDir(ctx context.Context, root string) (hash.Hash, error) {
...
+	}
+	if written != info.Size() {
+		return fmt.Errorf("%q changed size while hashing: expected %d bytes, read %d", path, info.Size(), written)
+	}
+	return nil
+}
+
+func writeFramedDirectoryBytes(h hash.Hash, value []byte) {
+	writeDirectoryUint64(h, uint64(len(value)))
+	_, _ = h.Write(value)
```
sbalabanov: make a comment why error is not processed. Same for below.

core/targethasher/sourcehasher.go
```diff
@@ -150,31 +152,135 @@ func hashDir(ctx context.Context, root string) (hash.Hash, error) {
...
+				return nil, err
+			}
+		}
+
+		info, err := os.Lstat(entry.path)
+		if err != nil {
+			return nil, err
+		}
+
+		writeFramedDirectoryBytes(dirHash, []byte(entry.relativePath))
```
sbalabanov: no need for relative path. The hash is content-only.

core/targethasher/sourcehasher.go
sbalabanov: Documented the intentionally ignored write results in d561d1dec9709f3615c0b2a748e70ca9fd1f7ee3. Directory writes now flow through one helper explaining that hash.Hash.Write is specified to consume all bytes without returning an error.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Removed the directory-level file-size field in d561d1dec9709f3615c0b2a748e70ca9fd1f7ee3. Each regular file now contributes its fixed-size content hash, making separate size framing redundant.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Eliminated raw filesystem-mode encoding in d561d1dec9709f3615c0b2a748e70ca9fd1f7ee3. The hash uses only platform-agnostic directory, regular-file, and symlink markers and rejects unsupported entry kinds.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Omitted permission and special mode bits in d561d1dec9709f3615c0b2a748e70ca9fd1f7ee3. The code now documents that modes are platform- and checkout-specific while entry kinds remain explicitly encoded.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Removed directory-hash versioning in d561d1dec9709f3615c0b2a748e70ca9fd1f7ee3. Backward compatibility is not required, so the stream now starts directly with deterministic entry records.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Removed absolute and root-relative paths from the hash in d561d1dec9709f3615c0b2a748e70ca9fd1f7ee3. Only framed local entry names remain, preserving BUG-007 rename and structure sensitivity without tying the hash to a checkout location.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov: Removed the repository-wide intermediate entry slice in d561d1dec9709f3615c0b2a748e70ca9fd1f7ee3. Hashing now happens directly in the WalkDir callback, whose lexical traversal keeps the result deterministic without retaining the whole tree.

[addressed by agent]

core/targethasher/sourcehasher.go
sbalabanov, line 160: add a doc explaining the function in one paragraph

sbalabanov: Added a concise Go doc for `hashDir` in 91e131904e21e38b0025bfef2870c33b6cb413e5.

The comment captures deterministic lexical traversal, streaming without whole-tree buffering, entry identity and payload semantics, and the intentional exclusion of checkout location and mode bits. This is documentation-only; behavior is unchanged.

[addressed by agent]

sbalabanov (commented):
Update PR description to correctly represent changes on the last version vs base revision. Make accent on edge cases it aims to address. Shorten. Use simple language.

sbalabanov:
Updated the BUG-007 commit message and PR description in fd67123a07dba5ca9b40c3810eff69e161ce963e to match the current two-file diff. The shorter text now highlights directory renames and structure, symlink target changes, deterministic streaming traversal, and the exclusion of checkout paths and mode bits. No code changed.

[addressed by agent]

core/targethasher/sourcehasher.go
```diff
@@ -145,30 +152,56 @@ func hashFile(path string) (hash.Hash, error) {
...
 
-		if d.Type().IsRegular() {
-			if fileCount%cancelCheckInterval == 0 {
-				if err := context.Cause(ctx); err != nil {
-					return err
-				}
+		switch {
+		case entry.Type()&fs.ModeSymlink != 0:
+			writeDirectoryEntryHeader(dirHash, symlinkEntryMarker, entry.Name())
+			target, err := os.Readlink(path)
```
xytan0056, line 184: would it be  an issue  if abosolute paths are difference across host? like `bazel info output_base` is different depending on where the query runs, and the files are actually the same?
maybe `fileHash.Sum` is enough, cause doesn't matter if file content is the same but paths are different

sbalabanov: Confirmed and fixed in 1c553e48b2affa0be99ef98c47a1841ab51515ab. Regular file and directory location paths are used only for I/O and never enter the digest. Absolute symlink targets were the exception; they are now canonicalized relative to the link before hashing, so identical relocated output trees hash the same while different logical targets remain distinct. Focused Go and Bazel tests pass.

[addressed by agent]

xytan0056 (approved):

PR #303, opened 2026-08-26
fix(itg): restore failed incremental hashes for BUG-029

sbalabanov:
## Summary

BUG-029 fixes incremental hash retries after a dependency hash fails. Hashes are now staged during traversal and written to the graph only after the full traversal succeeds; cycle detection uses separate temporary state.

## Use Case

For example, `//pkg:root` depends on a valid target and a source whose hash is temporarily unavailable. The first traversal fails. After the source hash becomes available, retrying on the same graph now computes the complete root hash instead of reusing partial state left by the failed attempt.

## Test Plan

- Added coverage for a failed dependency hash, cache serialization, and a successful retry on the same graph.
- Existing non-license CI checks pass.

## Revert Plan

Revert the single commit if incremental hash behavior regresses.

## Issues

T3-BUG-029

### DIFF
--- a/core/itg/graph/update.go
+++ b/core/itg/graph/update.go
@@ -186,6 +186,23 @@ func computeAvailableHashes(
 
 // computeHashes computes hashes recursively for the given target ID.
 func (g *OptimizedGraph) computeHashes(ctx context.Context, id int) ([]byte, error) {
+	computed := make(map[int][]byte)
+	hash, err := g.computeHashesRecursively(ctx, id, NewIntSet(), computed)
+	if err != nil {
+		return nil, err
+	}
+	for targetID, targetHash := range computed {
+		g.OptimizedTargets[targetID].Hash = targetHash
+	}
+	return hash, nil
+}
+
+func (g *OptimizedGraph) computeHashesRecursively(
+	ctx context.Context,
+	id int,
+	visiting IntSet,
+	computed map[int][]byte,
+) ([]byte, error) {
 	if ctx.Err() != nil {
 		return nil, context.Cause(ctx)
 	}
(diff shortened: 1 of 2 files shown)


sbalabanov (commented):
Simplify PR description, include use case example of the bug the pR is fixing

sbalabanov:
Updated the PR description to concisely explain BUG-029 and add a concrete failed-dependency retry use case. Code is unchanged; the current HEAD remains `976b5e3f05dfd19d0673c0716fb740297c74a994`.

[addressed by agent]

sbalabanov (commented):
Explain why you think the fix is necessary. What it means for the hash to be "temporary unavailable"? The algorithm does depth-first traversal to build a Merkle DAG. At some point the node is inspected and the hash updated. In case of error we do not retry on the same graph.
Prove me wrong.

sbalabanov:
I traced the full lifecycle and cannot prove BUG-029 is reachable in the current repository; your objection is correct.

- `OptimizedGraph.UpdateGraph` has no callers, and no production package imports `core/itg/graph` (only the unused ITG cache package does).
- In `UpdateGraph`, source hashes are completed before graph update/DFS. A source-hash error returns before `computeHashes`; missing files already receive deterministic hashes. Once DFS starts, normal graph construction supplies source hashes and rule hashes.
- A mid-DFS error can leave the old `[]byte{}` sentinel, but `UpdateGraph` returns immediately. There is no caller that retries or caches that errored graph.
- The added test bypasses this lifecycle: it directly constructs a source node with a nil hash, calls the private helper, manually supplies a hash, and retries. Its gob round trip is only asserted; the restored graph is never retried. It demonstrates the low-level mutation, not a reachable bug.

No code or metadata was changed. Draft status and HEAD `976b5e3f05dfd19d0673c0716fb740297c74a994` are unchanged. Closing or abandoning this speculative fix needs explicit instruction.

[addressed by agent]

PR #304, opened 2026-08-26
fix(itg): stabilize cycle hashing for BUG-028

sbalabanov:
## Summary

Fix BUG-028 by making incremental graph hashing deterministic when invalidated targets contain dependency cycles.

## Use case

A graph contains a cycle between `//pkg:a` and `//pkg:b`, with `//pkg:root` depending on the cycle. After a target changes, the old incremental update could traverse invalidated targets in Go map order and enter the cycle at a different target than a clean full recomputation. That produced different hashes for the same graph, leading to inconsistent incremental results and cache keys. This change hashes invalidated targets in canonical target-name order so both paths agree.

## Test Plan

- Run `bazel test //core/itg/graph:go_default_test`.
- Property-based tests permute invalidation order for rooted and rootless cycles.

## Revert Plan

Revert this PR.

## Issues

BUG-028

### DIFF
--- a/core/itg/graph/update_test.go
+++ b/core/itg/graph/update_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/uber/tango/core/targethasher"
+	"pgregory.net/rapid"
 )
 
 // fakeSourceHasher is a test double for targethasher.SourceHasher.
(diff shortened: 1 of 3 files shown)


sbalabanov:
simplify description and describe a use case showing a bug it fixes

sbalabanov:
Updated the BUG-028 description to focus on deterministic incremental cycle hashing and added a concrete rooted-cycle invalidation use case showing how the previous behavior could differ from a full recomputation. Current HEAD: `acb84d8e0e2a7c53e1dd4ddbdc17f539da043f22`.

[addressed by agent]

core/itg/graph/update_test.go
```diff
@@ -17,12 +17,14 @@ package graph
...
 	"context"
 	"crypto/sha1"
+	"slices"
 	"testing"
 
 	buildpb "github.com/bazelbuild/buildtools/build_proto"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/uber/tango/core/targethasher"
+	"pgregory.net/rapid"
```
sbalabanov, line 26: explain why this is needed. Is it used anywhere else in the repository?

sbalabanov: Current HEAD acb84d8e0e2a7c53e1dd4ddbdc17f539da043f22 uses this import for rapid.Check and rapid.Permutation in TestComputeInvalidatedHashesCycleOrderInvariance. Rapid generates many invalidation-order permutations and shrinks a failure to a minimal reproducer, so this test verifies the fix is independent of traversal order instead of covering only a few hand-picked orders. Yes, pgregory.net/rapid is already used by tests under core/cachekey, core/git, core/storage/disk, core/targethasher, core/workspace/githubpr, and internal/tgbdiff; it is also already declared in go.mod and MODULE.bazel.

[addressed by agent]

core/itg/graph/update_test.go
```diff
@@ -232,3 +234,100 @@ func TestComputeHashes(t *testing.T) {
...
+			rapid.Check(t, func(rt *rapid.T) {
+				graph := OptimizeGraph(tt.newTargets())
+				invalidated := NewIntSet()
+				for _, name := range rapid.Permutation(names).Draw(rt, "invalidation-order") {
+					invalidated.Insert(graph.TargetNameToID[name])
+				}
+
+				err := graph.computeInvalidatedHashes(ctx, invalidated)
+				if err != nil {
+					rt.Fatalf("compute invalidated hashes: %v", err)
```
sbalabanov: why cannot just assert on error? Is ir "rapid" limitation?

core/itg/graph/update_test.go
```diff
@@ -232,3 +234,100 @@ func TestComputeHashes(t *testing.T) {
...
+				}
+				for name, expected := range expectedHashes {
+					actual := graph.OptimizedTargets[graph.TargetNameToID[name]].Hash
+					if !slices.Equal(expected, actual) {
```
sbalabanov: assert?

core/itg/graph/update_test.go
sbalabanov: Updated in bdfc725003e5c53add92a58058558aed321e1168 to use require.NoError(rt, err) directly. rapid.T implements the testing methods required by Testify, and this repository already uses that pattern inside rapid.Check. require is preferable to assert here because a hashing error leaves no valid result to compare, while preserving Rapid failure reporting and shrinking behavior.

[addressed by agent]

core/itg/graph/update_test.go
sbalabanov: Updated in bdfc725003e5c53add92a58058558aed321e1168 to use assert.Equal(rt, expected, actual, ...) directly and removed the now-unused slices import. rapid.T supports Testify assertions, so a mismatch remains tied to the current generated example and Rapid can shrink it normally.

[addressed by agent]

xytan0056 (approved):

PR #306, opened 2026-08-27
perf(controller): eliminate redundant treehash storage reads

Arjunmehta312:
## Summary

Avoid rereading treehash values during `GetChangedTargets` by reading them
once and reusing them across cache lookup, graph lookup, and background
cache population.

## Validation

- Treehash storage reads reduced from 6 to 2 in the affected cache-miss
  path.
- Total storage reads reduced from 9 to 5 in the regression scenario.
- Added a regression test covering the redundant-read case.
- `make test` passes.
- `make lint` passes.

### DIFF
--- a/controller/gettargetgraph.go
+++ b/controller/gettargetgraph.go
@@ -56,7 +56,7 @@ func (c *controller) GetTargetGraph(request *pb.GetTargetGraphRequest, stream pb
 	if err != nil {
 		return tangoerrors.NewUser(fmt.Errorf("convert get target graph request: %w", err))
 	}
-	graphReader, err := c.getGraph(ctx, e, entityReq, "")
+	graphReader, err := c.getGraph(ctx, e, entityReq, nil, request.GetBuildDescription())
 	if err != nil {
 		return fmt.Errorf("get graph: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


xytan0056 (commented):
.

xytan0056 (approved):
Thanks for contributing. good catch

controller/gettargetgraph.go
```diff
@@ -96,34 +96,46 @@ func (c *controller) GetTargetGraph(request *pb.GetTargetGraphRequest, stream pb
 // entries store the full payload and stripping happens at send time, so
 // letting an orchestrator see it could poison the shared cache with
 // stripped graphs.
-func (c *controller) getGraph(ctx context.Context, e *metrics.Emitter, req entity.GetTargetGraphRequest) (storage.GraphReader, error) {
+// If treehash is provided (non-empty), it is used instead of reading from storage.
```
xytan0056: threading `""` as "not provided" through 4 function signatures is fragile — it conflates "caller didn't look it up" with "looked up, doesn't exist in storage."

Consider a separate resolver with memoization instead :

```go
type treehashResolver struct {
    storage storage.Storage
    emitter *metrics.Emitter
    op      string
    mu      sync.Mutex
    cache   map[string]resolvedTreehash
}

type resolvedTreehash struct {
    value string
    err   error
}

func (r *treehashResolver) resolve(ctx context.Context, build entity.BuildDescription) (string, error) {
    key := cachekey.GetTreehashCachePath(build)
    r.mu.Lock()
    if entry, ok := r.cache[key]; ok {
        r.mu.Unlock()
        return entry.value, entry.err
    }
    r.mu.Unlock()
    value, err := readTreehash(ctx, r.storage, build, r.emitter, r.op)
    r.mu.Lock()
    r.cache[key] = resolvedTreehash{value: value, err: err}
    r.mu.Unlock()
    return value, err
}
```

Create it at the top of `GetChangedTargets` and let `serveChangedTargetsFromCache`, `getGraph`, and `cacheComparedTargets` call `resolver.resolve()` directly. Second call for the same build is a map hit ,  no `""` sentinel, no conditional branches, and if a new code path needs the treehash later it just calls `resolve()` for free.

Arjunmehta312: Implemented this approach in the latest commit: replaced the `""` sentinel parameters with a request-scoped `treehashResolver`, which memoizes successful resolutions without caching errors. 
Thanks for the suggestion!

Arjunmehta312:
Rebased onto latest upstream/main (resolved conflict in `controller/getchangedtargets.go`).

**What was done:**
- Resolved merge conflict in `controller/getchangedtargets.go` (upstream added `maps` import, our branch needed `sync` - kept both)
- Rebased onto latest upstream/main (includes 10+ new commits from @sbalabanov-zz for BUG fixes)
- All 29 tests pass, race detector passes

**Implementation summary:**
- Added `treehashResolver` with memoization (successful values only, no error caching)
- `resolveParallel()` preserves concurrent reads + sibling cancellation
- Replaced `""` sentinel params with request-scoped resolver
- Treehash reads: 6 → 2 (66% reduction), total storage reads: 9 → 5

**Test results:** All 29 tests pass, race detector passes.

Ready for re-review @xytan0056

Arjunmehta312:
@xytan0056  Thanks again for the review
I’ve addressed your suggestion and preserved the parallel treehash resolution as well.
Whenever you get a chance, I’d appreciate a re-review.

PR #314, opened 2026-09-17
docs(rfc): propose async changed-targets API

behinddwalls:
GetChangedTargets holds a streaming RPC open through Bazel, so a timeout or disconnect aborts work with no handle to resume. This RFC adds ComputeChangeTargets, status, fetch, and cancel: persist the request, publish to a work queue, then return; a blocking consumer runs the existing two-graph compare path. RPC success requires queue publish; retries mint a new request_id. Native in-process Bazel is not a SubmitQueue build-signal poll loop.

### Why?

A client timeout or disconnect cancels request-bound Bazel with no durable handle, so retries repeat expensive work. The compared-target result is already a content-addressed blob, so computation can outlive the RPC.

### What?

Adds an illustrative async contract: ComputeChangeTargets stores a request and returns only after publishing `request_id`; GetChangeTargetsStatus and GetComputedChangedTargets are fetch-only; Cancel CAS-marks CANCELLED without killing in-flight Bazel. The consumer blocks on today's GetChangedTargets pipeline (parallel GetTargetGraph plus compare) and extends queue visibility until Process returns. No idempotency key: a retry is a new store row and publish; orphan rows after a failed publish are junk. SubmitQueue platform packages are reused for queue, consumer, counter, and DLQ (extract commons vs import submitqueue left open).

Co-authored-by: Cursor <cursoragent@cursor.com>


xytan0056 (changes requested):
Thanks for proposing! Stay tuned, we already have a feature planned. Callers should be  fairly simple without having the client to orchestrate everything. Will circle back

liujsj (changes requested):
Why is this an RFC

liujsj (changes requested):
No need for an RFC

